PR #3, opened 2020-03-09
Authentication service

nmamro:


AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Returns list of existing tokens (just an example)
+        /// </summary>
+        [HttpGet]
+        public IEnumerable<string> Get()
+        {
+            var rng = new Random();
+            return Enumerable.Range(0, tokens.Count - 1).Select(index => tokens[index])
+            .ToArray();
```
studokim: Why do we need to convert to Array?

DanilFadeeev (approved):

Maksim-Trolina (approved):

denis-kriv (approved):

Services/AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -0,0 +1,120 @@
...
+
+
+        [Route("[controller]/delete")]
+        [HttpGet]
+        public void DeleteToken(string token)
+        {
+            int index = BinarySearch(token);
+            TokensStorage.Tokens.RemoveAt(index);
+        }
+        private int BinarySearch(string token)
```
nmamro: Better to implement search in another class

Services/AuthentificationService/TokensStorage.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace AuthentificationService
+{
+    public static class TokensStorage
+    {
+        public static List<string> Tokens = new List<string>() { null };
```
nmamro: What's the reason of creating special class just for List? May be then it would be good to also add here methods like Add()/Remove() which we will just update as we change how tokens are stored?

Services/AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -0,0 +1,120 @@
...
+namespace AuthentificationService.Controllers
+{
+    [ApiController]
+    public class AuthentificationController : ControllerBase
+    {
+        /// <summary>
+        /// Storage for tokens 
+        /// May be used until database is created
+        /// </summary>
+
```
nmamro: Field was removed, why not to remove a comment then xd

Services/AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -66,6 +72,49 @@ public string GetToken()
...
         public string CheckToken(string token)
-            => tokens.Contains(token).ToString();
+        {
+            var index = BinarySearch(token);
+            int compare = String.Compare(TokensStorage.Tokens[index], token);
+            if (compare != 0)
+            {
+                return "false";
+            }
+            return "true";
```
studokim: How about `return compare == 0;`? Why do we need to return `string` instead of `bool`?

Services/AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        /// <summary>
+        /// Storage for tokens 
+        /// May be used until database is created
+        /// </summary>
+
+        /// <summary>
+        /// Constructor for lulz and some data to play with
+        /// </summary>
+       /* public AuthentificationController()
```
nmamro: If we don't use it at all may be it's better to remove it rather than adding /**/

Services/AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        /// <summary>
+        /// Generates and returns an active token
+        /// </summary>
+        [Route("[controller]/get")]
+        [HttpGet]
+        public string GetToken()
+        {
+            var date = DateTime.Now;
+            var randomDigit = new Random();
```
nmamro: Digit or number?

Services/AuthentificationService/TokensStorage.cs
Maksim-Trolina: If you think this is right, then you can remake

Services/AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -0,0 +1,120 @@
...
+            var valueForToken1 = randomDigit.Next(-400000000, 400000000);
+            var valueForToken2 = randomDigit.Next(-60000, 60000);
+            var valueForToken3 = randomDigit.Next(-60000, 60000);
+            var valueForToken4 = randomDigit.Next(-200, 200);
+            var byteArrayForGuid = new byte[]
+            {
+                (byte)date.Day,
+                (byte)date.Month,
+                (byte)(date.Year % 100),
+                (byte)(date.Year / 100),
```
nmamro: Why not something like `BitConverter.GetBytes(DateTime.UtcNow.ToBinary())` instead of this all that stuff?

Services/AuthentificationService/Controllers/AuthentificationController.cs
Maksim-Trolina: digit

Services/AuthentificationService/Controllers/AuthentificationController.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                return "false";
+            }
+            return "true";
+        }
+
+
+        [Route("[controller]/delete")]
+        [HttpGet]
+        public void DeleteToken(string token)
```
nmamro: If other methods are commented, why this one is not?

Services/AuthentificationService/Controllers/AuthentificationController.cs
studokim: Is this method going to work at all?

studokim (changes requested):

Services/AuthentificationService/Controllers/AuthentificationController.cs
Maksim-Trolina: > How about `return compare == 0;`? Why do we need to return `string` instead of `bool`?


This was done in order to test the operation of methods

Maksim-Trolina: > Is this method going to work at all?

It correctly determines if a token exists

dTarasova (approved):

studokim (approved):

AuthentificationService/Controllers/AuthentificationController.cs
LK777LK: See to the name convention

PR #12, opened 2020-03-17
Feature/user service check token

ifelseelif:


Services/UserService/Utils/TokenMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using UserService.Properties;
+
+namespace UserService.Utils
+{
+    public class TokenMiddleware
+    {
+        private RequestDelegate next;
```
dTarasova: private variables should be named with camelCase

Services/UserService/Utils/TokenMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public TokenMiddleware(RequestDelegate next)
+        {
+            this.next = next;
+        }
+        
+
+        public async Task InvokeAsync(HttpContext context)
+        {
+            if (context.Request.Method == "POST" & string.Compare(context.Request.Path, StringCons.UrlCreateUser, StringComparison.OrdinalIgnoreCase) == 0)
```
dTarasova: Not convenient way to read, it is better to wrap text in two lines.
Also, I think it is better to decompose such expressions because I don't clearly understand what and why do you compare and check

Services/UserService/Utils/TokenMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+            string userId = context.Request.Query["userId"];
+
+            if (checkToken(userId, token))
+            {
+                await next.Invoke(context);
+            }
+            else
+            {
+                context.Response.StatusCode = 403;
+                await context.Response.WriteAsync(StringCons.TokenError);
```
dTarasova: StringCons isn't understandable name. Also, it doesn't have a clear summary. I don't get what and why it is needed

dTarasova (approved):

Cerberous007 (approved):
Next time, I think you should add more comments

Services/UserService/Properties/StringConstants.resx
```diff
@@ -0,0 +1,129 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
DarkFelas: Not sure for this name, it is not constants class or something like it.
It is resources file, please take a look at sample name in Kernel project

Services/UserService/Utils/TokenMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    public class TokenMiddleware
+    {
+        private RequestDelegate next;
+
+        public TokenMiddleware(RequestDelegate next)
+        {
+            this.next = next;
+        }
+
```
DarkFelas: remove empty line

Services/UserService/Utils/TokenMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                try
+                {
+                    ans = client
+                        .GetStringAsync($"https://localhost:5001/Authentication/check?userId={userId}&token={token}")
+                        .Result;
+                }
+                catch (Exception)
+                {
+                    ans = null;
```
DarkFelas, line 56: It is a good idea to log this exception in future.
Mark this line with // TODO add logs

DarkFelas (changes requested):

Services/UserService/Utils/TokenMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+        public TokenMiddleware(RequestDelegate next)
+        {
+            this.next = next;
+        }
+        
+
+        public async Task InvokeAsync(HttpContext context)
+        {
+            if (context.Request.Method == "POST" 
+                & context.Request.Path == StringConstants.UrlCreateUser)
```
DarkFelas: Please read about difference between & and &&. in this case we need && and you will tell me why on next session

DarkFelas (changes requested):

Services/UserService/Utils/TokenMiddleware.cs
ifelseelif: fixed

DarkFelas (approved):

nmamro (approved):

Services/UserService/Utils/TokenMiddleware.cs
ifelseelif: fixed

Dimentr01 (approved):

PR #16, opened 2020-03-18
Feature/broker service tests and operation service tests

Maksim-Trolina:


DarkFelas:
One feature == one branch

PR #23, opened 2020-03-21
Feature/create broker service tests

Maksim-Trolina:


Tests/BrokerServiceTests/UnitTest1.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        [SetUp]
+        public void Setup()
+        {
+        }
+
+        [Test]
+        public void Test1()
+        {
+            Assert.Pass();
```
nmamro, line 15: I am not sure that this template is good. We don't want _UnitTest1.cs_ and `poblic void Test1` in develop branch.

nmamro (dismissed):

Tests/BrokerServiceTests/UnitTest1.cs
Maksim-Trolina: > I am not sure that this template is good. We don't want _UnitTest1.cs_ and `poblic void Test1` in develop branch.

This is just an empty project. Naturally, in the future the name will be correct, but so far the testing logic has not been developed so it makes no sense to come up with a name just like that

DarkFelas: Agree with Maksim

PR #26, opened 2020-03-23
UserService: create new postman tests collection

Dimentr01:


nmamro:
Do we integrate postman tests here or should we keep them in postman? I just don't really understand that now.

Dimentr01:
> Do we integrate postman tests here or should we keep them in postman? I just don't really understand that now.

we keep all tests in postman's collection

nmamro (approved):

Services/UserService/Startup.cs
```diff
@@ -32,7 +32,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseHttpsRedirection();
 
             app.UseRouting();
-            app.UseMiddleware<TokenMiddleware>();
+            //app.UseMiddleware<TokenMiddleware>();
```
ifelseelif: I think that here it is necessary to add TODO because antoher person can forget about this comment.

nmamro:
Since we decided to edit folders' structure in a Poll, it's better to do it here

Services/UserService/Startup.cs
Dimentr01: > I think that here it is necessary to add TODO because antoher person can forget about this comment.

maybe

Dimentr01: > I think that here it is necessary to add TODO because antoher person can forget about this comment.

I did

ifelseelif (approved):

Cerberous007 (approved):

DanilFadeeev (approved):

studokim (approved):

PR #28, opened 2020-03-23
DTO_CreateEmailPasswordModel

Cerberous007:


Common/DTO/UserEmailPassword.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DTO
+{
+    class UserEmailPassword
+    {
+        public string Email { get; set; }
+        public string Password { get; set; }
```
strannick-rodion: I think should be only get to not spoil the data

strannick-rodion (changes requested):

Common/DTO/UserEmailPassword.cs
Cerberous007: I think, you right, because this model used in client

Common/DTO/UserEmailPassword.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DTO
+{
+    class UserEmailPassword
+    {
+        public string Email { get; }
```
nmamro, line 5: How we will create a parametrized instance of it since `new UserEmailPassword() { Email = "", Password = "" }` is blocked?

Common/DTO/UserEmailPassword.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DTO
+{
+    class UserEmailPassword
```
nmamro: Public/private/internal/partial?

Common/DTO/UserEmailPassword.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
```
nmamro: Do we use all of them?

nmamro (changes requested):

Maksim-Trolina (approved):

Common/DTO/UserEmailPassword.cs
Cerberous007: Add public constructor

strannick-rodion (approved):

Common/DTO/UserEmailPassword.cs
Cerberous007: I don`t think about this. Already changed.

Common/DTO/UserEmailPassword.cs
Cerberous007: Public, i think

Common/DTO/UserEmailPassword.cs
Cerberous007: No, we don`t use this

studokim (approved):

nmamro (approved):

ifelseelif (approved):

PR #32, opened 2020-03-26
GUI_SignUpPage

DanilFadeeev:


Maksim-Trolina (approved):

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
+﻿@page "/signup"
+@using GUIModels
+@using GUI.Scripts
+@inject NavigationManager NavigationManager
+
+    <div class="col-12">
+        <h3><b>Sign up</b></h3>
+        <hr />
+
+        <EditForm Model="@userInput" OnValidSubmit="@TrySignUp">
```
dTarasova, line 10: Why do we have thos name? All validation named as ValidateSomething. It is understandable that we will try to do it

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
+﻿@page "/signup"
+@using GUIModels
+@using GUI.Scripts
+@inject NavigationManager NavigationManager
+
+    <div class="col-12">
```
dTarasova, line 6: These names of classes are automatical? Can we change them to more understandable?

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
...
+        </EditForm>
+    </div>
+
+
+@code {
+    bool isErrorDisplayed = false;
+    string errorMessage;
+    public void TrySignUp()
+    {
+        var Status = SignUper.SignUp(userInput);
```
dTarasova: If we declare this variable in method it has to begin with small letter

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
...
+        if (Status == System.Net.HttpStatusCode.OK)
+            NavigationManager.NavigateTo("/signin");
+        else
+        {
+            userInput = new SignUpViewModel();
+            isErrorDisplayed = true;
+            errorMessage = "данный e-mail уже занят";
```
dTarasova: Why Russian?!

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    private SignUpViewModel userInput;
```
dTarasova: Why variables and methods in the mess? Don't we have an agreement to write firs fields, then methods

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
...
+
+    protected override void OnInitialized()
```
dTarasova, line 55: What does this mean? Do you override some vs methods?  Why it is protected?

Clients/GUI/Scripts/SignUper.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Net;
+using DTO;
+using System.IO;
+using System.Text.Json;
+
+namespace GUI.Scripts
+{
+    public class SignUper
+    {
+        public static HttpStatusCode SignUp(SignUpViewModel insert)
```
dTarasova: I heard that HttpStatusCode isn't  good method to return. It will be  return automatically. Probably, better to make it void?

Clients/GUI/Scripts/SignUper.cs
```diff
@@ -0,0 +1,28 @@
+﻿using GUI.GUIModels;
+using System.Net;
+using DTO;
+using System.IO;
+using System.Text.Json;
+
+namespace GUI.Scripts
+{
+    public class SignUper
```
dTarasova: Strange name. What does it mean? this -er suffix

Clients/GUI/Scripts/SignUper.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public static HttpStatusCode SignUp(SignUpViewModel insert)
+        {
+            var output = new UserEmailPassword(insert.Email, insert.Password);
```
dTarasova: If you have output , why don't you named parameters of method "input"? 
Also, it isn't obvious. What concrete insert do you gave to this method. Do I have always to see which type of this variable it is ? Words "insert" and "output " not understandable at all

Clients/GUI/Pages/SignUp.razor
dTarasova: Status and StatusCode are different things . You mislead people

Clients/GUI/Shared/NavMenu.razor
```diff
@@ -6,7 +6,13 @@
 </div>
 
 <div class="@NavMenuCssClass" @onclick="ToggleNavMenu">
-
+    <ul class="nav flex-column">
+        <li class="nav-item px-3">
+            <NavLink class="nav-link"href="signup">
+                <span class="oi oi-plus" aria-hidden="true"></span> Sign Up
```
dTarasova: veeeery strange name of class :)

Clients/GUI/ViewModels/SignUpViewModel.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace GUI.GUIModels
+{
+    public class SignUpViewModel
+    {
+        [Required(ErrorMessage = "Email is required")]
```
dTarasova: Can you please explain what are thses attributes for?

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
...
+            <ValidationSummary />
+            <br />
+
+            <input type="submit" class="form-control col-1 button btn-primary" value="Sign up" />
+
+        </EditForm>
+    </div>
+
+
+@code {
```
dTarasova, line 37: What is this code doing? It is like handler of the form? Can we replace it in another file?

dTarasova (changes requested):

Clients/GUI/Pages/SignUp.razor
nmamro: These are bootstrap styles

Clients/GUI/Pages/SignUp.razor
nmamro: If you mean `<DataAnnotationsValidator />`, it's .net stuff

Clients/GUI/Pages/SignUp.razor
nmamro: It's pre-made stuff

Clients/GUI/Shared/NavMenu.razor
nmamro: Which one?

Clients/GUI/ViewModels/SignUpViewModel.cs
nmamro: It's for validation in gui form, at least now we don't have fluent validation yet, mb then we will be able to replace this with fluent validation stuff

Clients/GUI/Pages/SignUp.razor
nmamro: It seems to be a common practice for pages in blazor

nmamro (changes requested):
Dasha mentioned some stuff to fix, for now everything else looks fine

Clients/GUI/Scripts/SignUper.cs
DanilFadeeev: i corrected that

Clients/GUI/Scripts/SignUper.cs
DanilFadeeev: drive - driver, buy -buyer, swim - swimmer, SignUp - SignUper

nmamro: According to this logic, mb SignUpper then?

Clients/GUI/Pages/SignUp.razor
DanilFadeeev: corrected

Clients/GUI/Pages/SignUp.razor
DanilFadeeev: now it is correct

Clients/GUI/ViewModels/SignUpViewModel.cs
nmamro: UPD: yes, it will be done via fluent validation when the validation task will be done

Clients/GUI/Scripts/SignUper.cs
DanilFadeeev: how do we know that something went wrong?

Clients/GUI/Scripts/SignUper.cs
DanilFadeeev: yes, you are right

Clients/GUI/Pages/SignUp.razor
DanilFadeeev: i corrected that

Clients/GUI/Scripts/SignUper.cs
```diff
@@ -0,0 +1,28 @@
...
+            request.Method = "POST";
+            string jsonOutput = JsonSerializer.Serialize(output);
+            request.ContentType = "application/json";
+            using (var dataStream = new StreamWriter(request.GetRequestStream()))
+            {
+                dataStream.Write(jsonOutput);
+            }
+
+            var response = (HttpWebResponse)request.GetResponse();
+            return response.StatusCode;
```
nmamro, line 31: We can do it like that: 
```
if (response.IsSuccessStatuscode)
{
// do sth 
}
else
{
//throw some exception
}
```
And then we will handle that in GUI code

nmamro: I'm not sure whether it's the best option, but in another pull request Spartak mentioned that returning `HttpStatusCode` is not a good practice

Clients/GUI/Scripts/SignUper.cs
dTarasova: Hm. You corrected just first line. Input and output also not so understandable. 

And still: why do you continue to return HttpStatusCode? Maybe it would be beter if we will make it also void and to throw exceptions here? Why do you have to check it in another method

ifelseelif (approved):

Dimentr01 (approved):

LK777LK (approved):

Clients/GUI/Scripts/SignUper.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace GUI.Scripts
+{
+    public class SignUpper
+    {
+        public static HttpStatusCode SignUp(SignUpViewModel input)
+        {
+            var output = new UserEmailPassword(input.Email, input.Password);
+
+            var request = (HttpWebRequest)WebRequest.Create("https://localhost:5011/user/create");
```
studokim, line 21: Is there a way to avoid hardcoding the url?

Clients/GUI/ViewModels/SignUpViewModel.cs
```diff
@@ -0,0 +1,16 @@
...
+        public string Email { get; set; }
+
+        [Required(ErrorMessage = "Password is required")]
+        public string Password { get; set; }
+
+        [Compare(nameof(Password))]
+        public string ConfirmPassword { get; set; }
+    }
+}
```
studokim: Do we need two identical classes?

Common/DTO/UserEmailPassword.cs
```diff
@@ -2,8 +2,9 @@
 {
     public class UserEmailPassword
     {
-        public string Email { get; }
-        public string Password { get; }
+        UserEmailPassword() { }
```
studokim: According to convention, private constructors must be placed after public ones

Common/HttpWebRequestWrapperLib/HttpWebRequestWrapper.cs
```diff
@@ -0,0 +1,194 @@
...
+        }
+        
+        public string Get(string url, Dictionary<string, string> queryParams = null,
+            Dictionary<string, string> headerCollection = null, Dictionary<string, string> cookieContainer = null)
+        {
+            if (string.IsNullOrEmpty(url))
+            {
+                throw new NullReferenceException();
+            }
+            url = getUrlWithParams(url, queryParams);
```
studokim: NullReference is checked in this method. Do we need the another "if" above?

Common/HttpWebRequestWrapperLib/HttpWebRequestWrapper.cs
```diff
@@ -0,0 +1,194 @@
...
+            httpWebRequest.Method = "GET";
+            httpWebRequest.Headers = getHeaderCollectionFromDictionary(headerCollection);
+            httpWebRequest.CookieContainer = getCookieContainerFromDictionary(cookieContainer);
+
+            var httpWebResponse = (HttpWebResponse)httpWebRequest.GetResponse();
+
+            // TODO: Handling custom exceptions
+            if (httpWebResponse.StatusCode != HttpStatusCode.OK)
+            {
+                throw new Exception("Response returned with error");
```
studokim: How about specializing the Exception?

strannick-rodion (commented):

Clients/GUI/ViewModels/SignUpData.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace GUI.GUIModels
+{
+    public class SignUpData
```
nmamro, line 5: Seems like we don't use this class anymore

Clients/GUI/Scripts/SignUper.cs
dTarasova: What is hardcoding mean?

Cerberous007: In this case, it is an explicit url

nmamro:
This request is active for kinda long time, better to make changes a bit faster

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
...
+    public void TrySignUp()
+    {
+        var statusCode = SignUpper.SignUp(userInput);
+        if (statusCode == System.Net.HttpStatusCode.OK)
+            NavigationManager.NavigateTo("/signin");
+        else
+        {
+            userInput = new SignUpViewModel();
+            isErrorDisplayed = true;
+            errorMessage = "this e-mail already taken";
```
Cerberous007, line 51: I think this error мessage is not correct. Email can be free, but registration can be failed for another reason. 
For example, connection between services was lost or database was fallen and etc.

Clients/GUI/Scripts/SignUper.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.IO;
+using System.Text;
+using System.Text.Json;
+
+namespace GUI.Scripts
+{
+    public class SignUpper
+    {
+        public static HttpStatusCode SignUp(SignUpViewModel input)
+        {
```
Cerberous007, line 18: Now you can use HttpWebRequestWrapperLib instead this stuff

Common/HttpWebRequestWrapperLib/HttpWebRequestWrapper.cs
```diff
@@ -0,0 +1,194 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using System.Text;
+using System.Text.Json;
+
+namespace HttpWebRequestWrapperLib
+{
```
Cerberous007: I don`t now, how it got here, but we already have HttpWebRequestWrapper.
I think you should merge new develop into your branch and take en existing library from there.

Cerberous007 (changes requested):

Maksim-Trolina (approved):

Clients/GUI/Pages/SignUp.razor
DenisKrivoruchko11: why go to the signin page and not on the userinfo page

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,59 @@
...
+@code {
+    bool isErrorDisplayed = false;
+    string errorMessage;
+    private SignUpViewModel userInput;
+
+    public void TrySignUp()
+    {
+        var statusCode = SignUpper.SignUp(userInput);
+        if (statusCode == System.Net.HttpStatusCode.OK)
+            NavigationManager.NavigateTo("/signin");
```
denis-kriv, line 46: why do we need go on signin page, why not on userinfo page?

denis-kriv (changes requested):

Clients/GUI/Pages/SignUp.razor
Cerberous007: Because user has just registered. If he wants to go to userinfo page, he must log in

PR #33, opened 2020-03-27
AuthService_UserEmailPasswordModelValidation

nmamro:


Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
```diff
@@ -0,0 +1,25 @@
...
+            RuleFor(user => user.Email)
+                .NotEmpty()
+                .WithMessage("Email address must not be empty.")
+                .EmailAddress()
+                .WithMessage("Email address must be in valid format.");
+
+            RuleFor(user => user.Password)
+                .NotEmpty()
+                .WithMessage("Password must not be empty.")
+                .Length(5, 500)
```
DarkFelas: 500 chars password oO
Not bad, not bad

DanilFadeeev (approved):

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
```diff
@@ -0,0 +1,25 @@
+﻿using DTO;
+using FluentValidation;
+
+namespace AuthenticationService.Validators
```
dTarasova, line 4: Do we have to have this stuff only in authentification service? 
Why we don't move this to common validation? For example , we have to check email and password in registration also

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
```diff
@@ -0,0 +1,25 @@
...
+                .WithMessage("Email address must not be empty.")
+                .EmailAddress()
+                .WithMessage("Email address must be in valid format.");
+
+            RuleFor(user => user.Password)
+                .NotEmpty()
+                .WithMessage("Password must not be empty.")
+                .Length(5, 500)
+                .WithMessage("Password's length must be from 5 to 500 symbols.")
+                .Matches(@"^[a-zA-Z0-9]+$")
```
dTarasova: And how about spaces, symbols? It is just a suggestion. 

Also, I think 500 is a little more that people needed

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
```diff
@@ -0,0 +1,25 @@
...
+                .WithMessage("Password must be alphanumeric.");
```
dTarasova: How about "Password should contain only letters and numbers"?

Tests/AuthServiceTests/Validators/UserEmailPasswordValidatorTests.cs
```diff
@@ -0,0 +1,99 @@
...
+{
+    public class UserEmailPasswordValidatorTests
+    {
+        private UserEmailPassword emailPasswordEmpty = new UserEmailPassword(string.Empty, string.Empty);
+
+        private UserEmailPassword emailEmptyPasswordOk = new UserEmailPassword(string.Empty, "xxxxx");
+
+        private UserEmailPassword passwordEmptyEmailOk = new UserEmailPassword("example@gmail.com", string.Empty);
+
+        private UserEmailPassword emailInvalidFormat = new UserEmailPassword("ololo", "xxxxx");
```
dTarasova: Why not emailInvalidFormatPasswordOk? Don't you have to write in the same style? 
If you check both of fields, check it everywhere. Oh you have to split it to 2 differents examples
And the same question to lines below

dTarasova (changes requested):

Dimentr01 (approved):

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: Why would anyone need password like _"  /-==/%/*[]^][ [ ]\ [  \"_, it's not cool :(

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: What would be the difference? Same meaning but more words

Tests/AuthServiceTests/Validators/UserEmailPasswordValidatorTests.cs
nmamro: Might have missed that

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: The task item was marked with authentication service, but i get your point, will think about it

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: I wanted to set 200-300 characters long password once, and when the service's response was like "password can't be longer than 50 symbols", I was a bit disappointed actually, because it would be fun otherwise

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: You can also suggest where you think this validator needs to be located

dTarasova: I think it would be okay in common. What if we create folder fluentvalidation in project CommonValidation

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
dTarasova: Wow, seriously? Okay :) 
Still don't understand why, but okay. 

Yes, that's password is bad, but how about "my password" or "kalinka-malinka" or sth like that

dTarasova: And also, I think that most of people have small passwords. Like.. 98 % will have password less than 50 symbols. And you will always have memory for 450 symbols

dTarasova: And do you have checks for length of email?

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
dTarasova: I think it is more friendly. But it was just a suggestion

LK777LK (approved):

ifelseelif (approved):

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
studokim: By the way, some people use password generators with special symbols. In our _trading platform_ a strong password is essential

strannick-rodion (commented):

Tests/AuthServiceTests/Validators/UserEmailPasswordValidatorTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using FluentValidation;
+using NUnit.Framework;
+using System.Linq;
+
+namespace AuthenticationServiceTests.Validators
+{
+    public class UserEmailPasswordValidatorTests
+    {
+        private UserEmailPassword emailPasswordEmpty = new UserEmailPassword(string.Empty, string.Empty);
+
```
Cerberous007, line 12: Why you don``t use  `var` here and below?

nmamro:
> You should to change name to follow our Code Convention

Very explicit review, thank you. May I know which name? :-)

Tests/AuthServiceTests/Validators/UserEmailPasswordValidatorTests.cs
nmamro: Because these are fields, not variables

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: > And do you have checks for length of email?

Whaaaat, where?

nmamro: > And also, I think that most of people have small passwords. Like.. 98 % will have password less than 50 symbols. And you will always have memory for 450 symbols

What? Validation != giving memory

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: Do we have a CommonValidation project? 0_0
As I can see, we have project Kernel with CommonValidation.cs class

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
nmamro: > By the way, some people use password generators with special symbols. In our _trading platform_ a strong password is essential

Well, agree, removed the limitation.

nmamro:
> Why 420?

My bad, accidentally forgot to change it to 50 everywhere

Services/AuthenticationService/Validators/UserEmailPasswordValidator.cs
dTarasova: Oh, I missed that. Okay, we can create in Kernel folder with Validation

dTarasova (approved):

PR #35, opened 2020-03-27
Feature/PostmanTest_TestsForNewsService

ifelseelif:


DanilFadeeev (approved):

dTarasova (approved):

Integrations/CentralBankIntegrationLib/XmlSerializationObjects/ValCurse.cs
```diff
@@ -13,6 +13,6 @@ public class ValCurs
         [XmlElement("Valute")]
         public List<Valute> Valutes { get; set; }
 
-        public List<string> TopValutesCharCode => new List<string> {  };
+        public List<string> TopValutesCharCode => new List<string> { "EUR", "USD", "GBR", "CNY", "JPY" };
```
LK777LK, line 16: We need this change because code this integration lib is not correct

LK777LK: Change back

LK777LK (changes requested):

Integrations/CentralBankIntegrationLib/XmlSerializationObjects/ValCurse.cs
ifelseelif: why is it not correct ? if I change the code, it will always output "No currencies found" in method  CentralBankNewsPublisher.GetNews

ifelseelif: I didn't find in the code that someone filled this list

studokim (approved):

strannick-rodion (commented):

strannick-rodion (approved):

Cerberous007 (approved):

LK777LK (approved):

PR #42, opened 2020-04-01
Common_CreateCustomException

Maksim-Trolina:


Common/CustomException/ForbiddenException.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
```
nmamro: Do we use all these usings?

Cerberous007 (approved):
Don`t forget to remove useless usings and merge develop in your branch before merge

DanilFadeeev (approved):

denis-kriv (approved):

dTarasova (changes requested):
Why don't we standardize messages in exceptions? 
I think we also should write in summary situations where we use these exceptions

Maksim-Trolina:
> Why don't we standardize messages in exceptions?
> I think we also should write in summary situations where we use these exceptions


At this stage, this is enough for us then we will add our logic there, if necessary

studokim (approved):
Agree with Dasha about summary, but okay

Common/CustomException/BadRequestException.cs
```diff
@@ -0,0 +1,18 @@
...
+    public class BadRequestException : Exception
+    {
+        public BadRequestException(string message)
+            : base(message)
+        { }
+
+        public BadRequestException()
+        { }
+
+        public BadRequestException(string message, Exception inner)
```
ifelseelif, line 14: why do we need this constructor with Exception inner ?

DarkFelas: Test slack integration

DarkFelas (changes requested):
Test slack

DarkFelas (approved):

Common/CustomException/BadRequestException.cs
ifelseelif: what does it mean? I didn't understand.

DarkFelas: Check telegram ))

ifelseelif (approved):

nmamro:
Seems ok overall for me, waiting for Dasha

Dimentr01 (approved):

PR #46, opened 2020-04-01
GUI_SignUpPage

DanilFadeeev:


Clients/GUI/Scripts/SignUpper.cs
```diff
@@ -0,0 +1,18 @@
+using GUI.GUIModels;
+using System.Net;
+using DTO;
+using System.IO;
```
studokim: Are all the Usings essential?

Clients/GUI/Scripts/SignUpper.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Text.Json;
+using HttpWebRequestWrapperLib;
+
+namespace GUI.Scripts
+{
+    public class SignUpper
```
studokim: Why the class isn't Static while its only method is?

Clients/GUI/Scripts/SignUpper.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public static void SignUp(SignUpViewModel input)
```
studokim: We already have `public static void SignIn(UserEmailPassword userInput)`. It's better to keep one type of arguments

Clients/GUI/ViewModels/SignUpViewModel.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.ComponentModel.DataAnnotations;
 
-namespace GUI.ViewModels
+namespace GUI.GUIModels
```
studokim: I suppose we need to rename the method(s) as well

studokim (changes requested):

Clients/GUI/Pages/SignUp.razor
```diff
@@ -0,0 +1,61 @@
...
+        try
+        {
+            SignUpper.SignUp(userInput);
+            NavigationManager.NavigateTo("/signin");
+        }
+        catch
+        {
+            userInput = new SignUpViewModel();
+            isErrorDisplayed = true;
+            errorMessage = "this e-mail already taken";
```
studokim, line 54: By the way, it was said in the previous PR: we may encounter a different error, so the message may be misleading

Clients/GUI/Scripts/SignUpper.cs
DanilFadeeev: first, we dont need take it in memory all time, second, if many people would like to sign up, static class cant service it all in one time, in contrast many different signuppers

Dimentr01 (approved):

Clients/GUI/ViewModels/SignUpViewModel.cs
DanilFadeeev: what name is your favorit?

studokim: Not favourite, but uniform. We've renamed the Namespace to _GUIModels_, so why methods are still _...ViewModel..._?

Clients/GUI/Scripts/SignUpper.cs
studokim: Its only method is static nevertheless. So it _is_ in memory all the time, and static method still can't have simultaneous instances of itself

Clients/GUI/Scripts/SignUpper.cs
ifelseelif: If this method, for example, takes n seconds, then other users will not be able to get a page with a log up about n seconds

ifelseelif (approved):

LK777LK (approved):

Clients/GUI/Shared/NavMenu.razor
```diff
@@ -7,15 +7,11 @@
 
 <div class="@NavMenuCssClass" @onclick="ToggleNavMenu">
     <ul class="nav flex-column">
-        <AuthorizeView>
-            <Authorized>
-                <li class="nav-item px-3">
```
zzzloba, line 12: Why you removed it?

Clients/GUI/ViewModels/SignUpViewModel.cs
nmamro: I have no idea why we renamed it to GUIModels, it should be ViewModels

nmamro: Fix it please

Clients/GUI/Scripts/SignUpper.cs
nmamro: Fixed

Clients/GUI/Scripts/SignUpper.cs
nmamro: Fixed

nmamro (approved):

studokim (approved):

PR #47, opened 2020-04-01
Test slack

DarkFelas:


Common/DTO/User.cs
```diff
@@ -4,9 +4,9 @@ namespace DTO
 {
     public class User
     {
-        public Guid Id { get; }
-        public string Login { get; }
+        public Guid Id { get; set; }
```
DarkFelas, line 7: WTF

DarkFelas (commented):
Need more work

Common/DTO/User.cs
DarkFelas: Test slack

PR #48, opened 2020-04-02
Common_CreateCustomException

Maksim-Trolina:


Common/CustomException/ForbiddenException.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace CustomException
+{
+    /// <summary>
+    /// Indicates that the server understood the request, but refuses to authorize it
```
Cerberous007, line 6: Not refuses to authorize
I think, there should be something like that: "user must be authorized"

ifelseelif (approved):

Cerberous007 (changes requested):

Common/CustomException/ForbiddenException.cs
Maksim-Trolina: > Not refuses to authorize
> I think, there should be something like that: "user must be authorized"


So the translator translated, but the meaning is clear

LK777LK (approved):

nmamro (approved):

nmamro:
Waiting for Dmitry

nmamro:
And for resolving conflicts

Common/CustomException/ForbiddenException.cs
Cerberous007: Ok, may be my english isn`t so well))

Cerberous007 (approved):

Common/CustomException/ForbiddenException.cs
strannick-rodion: May be you should use not authorize or execute

PR #49, opened 2020-04-02
DataBaseService_DeleteUserLogic

strannick-rodion:


LK777LK (approved):

LK777LK (approved):

Services/DataBaseService/DbModels/DbUserCredential.cs
```diff
@@ -9,6 +9,7 @@ public class DbUserCredential
         public Guid Id { get; set; }
         public string Email { get; set; }
         public string PasswordHash { get; set; }
+        public bool IsExist { get; set; }
```
nmamro: Exists mb?

Services/DataBaseService/Repositories/UserCredentialRepository.cs
```diff
@@ -31,6 +31,16 @@ public void Create(UserEmailPassword data)
                 throw new Exception(e.Message + "\nCan`t add user");       
             }
         }
+
+        public void Delete(Guid data)
+        {
+            var user = dbContext.Find<DbUserCredential>(data);
+            if (user!=null)
```
nmamro: Spaces in condition

Services/DataBaseService/Utils/MigrationScripts/24.03.2020/0002_AddIsExistColumnToUsersCredential.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE dbo.UsersCredentials ADD IsExist BIT DEFAULT 1;
```
nmamro: Again, I don't like the name

studokim: Agree. Moreover, IsActive sounds better for me since we can Deactivate the user by deleting him, and then Activate via hotline's admin panel

Services/DataBaseService/DbModels/DbUserCredential.cs
nmamro: Or IsActive (as Kirill said below)

Services/DataBaseService/Repositories/UserCredentialRepository.cs
studokim: `(!(user is null))` is recommended
https://www.gullberg.tk/blog/is-null-versus-null-in-c/

Services/DataBaseService/Interfaces/IRepository.cs
```diff
@@ -1,7 +1,10 @@
-﻿namespace DataBaseService.Interfaces
+﻿using System;
+
+namespace DataBaseService.Interfaces
 {
     public interface IRepository<Input>
     {
         void Create(Input data);
+        void Delete(Guid data);
```
studokim: How about
   ```
 public interface IRepository<Input, Id>
    {
        void Create(Input data);
        void Delete(Id data);
```
or something? Don't know if it's the right way, but the approach seems more flexible

File: Services/DataBaseService/DbModels/DbUserCredential.cs (hunk 1 above)
Comment (strannick-rodion) [reply Services/DataBaseService/DbModels/DbUserCredential.cs]: I have thought about it, but it can conflict with online user or active (when it execute other operation)

File: Services/DataBaseService/Utils/MigrationScripts/24.03.2020/0002_AddIsExistColumnToUsersCredential.sql (hunk 3 above)
Comment (strannick-rodion) [reply Services/DataBaseService/Utils/MigrationScripts/24.03.2020/0002_AddIsExistColumnToUsersCredential.sql]: I have answered abow

File: Services/DataBaseService/Interfaces/IRepository.cs (hunk 4 above)
Comment (strannick-rodion) [reply Services/DataBaseService/Interfaces/IRepository.cs]: I don't know too, but I think it isn't because, if we add something more we should  changed a  lot in everytime  when we will inherited from the repository

File: Services/DataBaseService/Repositories/UserCredentialRepository.cs (hunk 2 above)
Comment (strannick-rodion) [reply Services/DataBaseService/Repositories/UserCredentialRepository.cs]: I don' understand.Can you explain?

File: Services/DataBaseService/DbModels/DbUserCredential.cs (hunk 1 above)
Comment (nmamro) [reply Services/DataBaseService/DbModels/DbUserCredential.cs]: > I have thought about it, but it can conflict with online user or active (when it execute other operation)

Then come up with any other option which follows english grammar

Comment (nmamro) [reply Services/DataBaseService/DbModels/DbUserCredential.cs]: For example 'Exists' as I said above

**Review by Cerberous007 — APPROVED** [review]

File: Services/DataBaseService/Contollers/DataBaseController.cs [hunk 5]
```diff
@@ -1,18 +1,26 @@
...
 using DTO;
 using DataBaseService.Interfaces;
+using System;
 
 namespace DataBaseService.Contollers
 {
     [ApiController]
-    [Route("[controller]")]    
+    [Route("[controller]")]
     public class DataBaseController : ControllerBase
```
nmamro: Actually we won't need controller after merging Mass Transit, but it's k now, just telling

Services/DataBaseService/Repositories/UserCredentialRepository.cs
```diff
@@ -31,6 +31,16 @@ public void Create(UserEmailPassword data)
                 throw new Exception(e.Message + "\nCan`t add user");       
             }
         }
+
+        public void Delete(Guid data)
+        {
+            var user = dbContext.Find<DbUserCredential>(data);
```
nmamro: What happens when user is not found?

strannick-rodion: It returns null

Services/DataBaseService/Contollers/DataBaseController.cs
strannick-rodion: ок

studokim (approved):
Well, Exist**s**, but okay

Services/DataBaseService/Commands/DeleteUserCommand.cs
```diff
@@ -0,0 +1,22 @@
+﻿using DataBaseService.Interfaces;
+using DTO;
+using Microsoft.AspNetCore.Mvc;
+using System;
+
+namespace DataBaseService.Commands
+{
+    public class DeleteUserCommand : ICommand<Guid>
```
dTarasova, line 8: Isn't this make conflicts with create logic? 
Now we return id in order to check that user was created . But then they will be void both

Services/DataBaseService/DbModels/DbUserCredential.cs
```diff
@@ -9,6 +9,7 @@ public class DbUserCredential
         public Guid Id { get; set; }
         public string Email { get; set; }
         public string PasswordHash { get; set; }
+        public bool Exist { get; set; }
```
dTarasova: The word exist isn't understandable. I think that it means that user hadn't delete himself. But it is dbmodel. This User always exist in database

Services/DataBaseService/Interfaces/IRepository.cs
dTarasova: If you fix sth, you have to write "done" or "fix"
Not just resolve conversation

Services/DataBaseService/DbModels/DbUserCredential.cs
nmamro: Exists

nmamro (changes requested):

Services/DataBaseService/Repositories/UserCredentialRepository.cs
```diff
@@ -31,6 +31,16 @@ public void Create(UserEmailPassword data)
                 throw new Exception(e.Message + "\nCan`t add user");
```
dTarasova: Why don't custom exc?

dTarasova (changes requested):

Services/DataBaseService/DbModels/DbUserCredential.cs
strannick-rodion: It's true, but do you have more suit word?

Services/DataBaseService/Repositories/UserCredentialRepository.cs
strannick-rodion: It isn't part of code

Services/DataBaseService/Commands/DeleteUserCommand.cs
strannick-rodion: It works

Services/DataBaseService/DbModels/DbUserCredential.cs
strannick-rodion: In SQL scripts it is a special word

strannick-rodion: I can use IsExist or Exist? what way?

strannick-rodion: I use IsExists

Services/DataBaseService/Commands/CreateUserCommands.cs
```diff
@@ -0,0 +1,27 @@
...
+            this.userRepository = userRepository;
+        }
+
+        public void Execute(UserEmailPassword data)
+        {
+            //TODO Change to custom exception
+            if (string.IsNullOrEmpty(data.Email) || string.IsNullOrEmpty(data.PasswordHash))
+            {
+                throw new Exception("Not correct data");
+            }
```
dTarasova, line 23: Change to Custom Exception

Services/DataBaseService/Commands/DeleteUserCommand.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        private readonly IRepository<UserEmailPassword> userRepository;
+
+        public DeleteUserCommand([FromServices] IRepository<UserEmailPassword> userRepository)
+        {
+            this.userRepository = userRepository;
+        }
+
+        public void Execute(Guid data)
```
dTarasova, line 17: Here you can also change to id

Services/DataBaseService/Repositories/UserCredentialRepository.cs
```diff
@@ -32,6 +33,20 @@ public Guid Create(UserEmailPassword data)
...
             }
         }
+
+        public void Delete(Guid id)
```
dTarasova, line 37: Where is id null check?

Services/DataBaseService/Repositories/UserCredentialRepository.cs
```diff
@@ -32,6 +33,20 @@ public Guid Create(UserEmailPassword data)
...
+        {
+            var user = dbContext.Find<DbUserCredential>(id);
+            if (!(user is null))
```
dTarasova, line 40: How about switch these blocks of code : 
if (user is null) throw 
else user.IsExists = false; dbContext.SaveChanges();

ifelseelif: Why are you using this construct instead of  user ! = null ?

Services/DataBaseService/Commands/CreateUserCommands.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public CreateUserCommand([FromServices] IRepository<UserEmailPassword> userRepository)
+        {
+            this.userRepository = userRepository;
+        }
+
+        public void Execute(UserEmailPassword data)
+        {
+            //TODO Change to custom exception
+            if (string.IsNullOrEmpty(data.Email) || string.IsNullOrEmpty(data.PasswordHash))
```
ifelseelif, line 20: Will you replace it with fluent validation ?

Services/DataBaseService/Repositories/UserCredentialRepository.cs
studokim: > 
> 
> Why are you using this construct instead of user ! = null ?

https://www.gullberg.tk/blog/is-null-versus-null-in-c/

Services/DataBaseService/Commands/CreateUserCommands.cs
strannick-rodion: Not, we shouldn't it validates and this file was delete from develop and I return it

Services/DataBaseService/Commands/CreateUserCommands.cs
strannick-rodion: this file was delete from develop and I return it

Services/DataBaseService/Commands/DeleteUserCommand.cs
strannick-rodion: I think it isn't necessary

Services/DataBaseService/Repositories/UserCredentialRepository.cs
strannick-rodion: It should check UserService

Services/DataBaseService/Repositories/UserCredentialRepository.cs
strannick-rodion: I will change it, if I change something else, but I think it is not principal

Services/DataBaseService/Commands/CreateUserCommands.cs
dTarasova: So? Why don't we make it better ?

Services/DataBaseService/Repositories/UserCredentialRepository.cs
dTarasova: Yeah. But we talked about this. What if user service will be rewrite? Why do we have to hope on other services? Also, I don't see something terrible in one more null check

Services/DataBaseService/Repositories/UserCredentialRepository.cs
dTarasova: It is only one commit and not very long one. And code will look better.

Services/DataBaseService/Repositories/UserCredentialRepository.cs
strannick-rodion: It is the task User Service and we trust UserService

Services/DataBaseService/Repositories/UserCredentialRepository.cs
strannick-rodion: ок

Services/DataBaseService/Commands/CreateUserCommands.cs
strannick-rodion: I won't change this code because it delete in develop now

DanilFadeeev (approved):

VLiamin (approved):

nmamro (changes requested):
Fix conflicts

ifelseelif (approved):

PR #51, opened 2020-04-03
SeveralServices_Integration of MassTrasnit for login and create

nmamro:


nmamro:
I changed logic of Create in repository to be able to return Id and login immediately after registration.
If you disagree, comment why.

studokim (approved):

Services/DataBaseService/BrokerConsumers/UserLoginConsumer.cs
```diff
@@ -0,0 +1,28 @@
+﻿using DTO;
```
DarkFelas, line 1: This class not needed right now, you can delete it

PR #53, opened 2020-04-03
GUI_Get/Save token and working logic of sign in/up

nmamro:


Services/AuthenticationService/Controllers/AuthenticationController.cs
```diff
@@ -31,7 +33,9 @@ public string Login([FromServices] ILoginCommand command, [FromBody] UserEmailPa
         {
             var result = command.Execute(user);
 
-            return result.Result;
+            var userToken = result.Result;
```
dTarasova: I suggest to change the name of result. What type it is? Why does result have result properties. It isn't understandable

Clients/GUI/Pages/SignIn.razor
```diff
@@ -35,13 +32,15 @@
...
-        var user = new UserEmailPassword(userInput.Email, userInput.Password);
         try
         {
-            // TODO: add logic after MassTransit flow iw merged
-            //SignInner.SignIn(user);
+            await LocalStorage.ClearAsync();
+            var hash = ShaHash.GetHash(userInput.Password);
+            var credentials = new UserEmailPassword(userInput.Email, hash);
+            var userToken = SignInner.SignIn(credentials);
+            LocalStorage.SetItemAsync(userToken.UserId.ToString(), userToken.Body);
```
dTarasova: Why github isn' satisfied?

Clients/GUI/Pages/SignUp.razor
```diff
@@ -1,57 +1,62 @@
...
+            var hash = ShaHash.GetHash(userInput.Password);
+            var credentials = new UserEmailPassword(userInput.Email, hash);
+            SignUpper.SignUp(credentials);
+            await LocalStorage.ClearAsync();
+            var userToken = SignInner.SignIn(credentials);
+            LocalStorage.SetItemAsync(userToken.UserId.ToString(), userToken.Body);
+            AuthenticationProvider.MarkSignedIn();
+            NavigationManager.NavigateTo("/userinfo");
         }
         catch
```
dTarasova: Don't we suppose to catch customs exceptions? and have different handling for them?

Clients/GUI/Scripts/SignUpper.cs
```diff
@@ -1,13 +1,18 @@
...
 {
     public static class SignUpper
     {
-        public static void SignUp(UserEmailPassword data)
+        public static Guid SignUp(UserEmailPassword data)
         {
-            new HttpWebRequestWrapper().Post("https://localhost:5011/user/create", null, data);
+            var wrapper = new HttpWebRequestWrapper();
+            var response = wrapper.Post("https://localhost:5011/user/create", null, data);
+            return JsonSerializer.Deserialize<Guid>(response);
```
dTarasova: I am not sure, but here we don't need try catch?

Clients/GUI/Shared/NavMenu.razor
```diff
@@ -52,6 +53,7 @@
 
     private void SignOut()
     {
+        LocalStorage.ClearAsync();
```
dTarasova, line 56: Local Storage is different object for each user?

Services/AuthenticationService/Commands/LoginCommand.cs
```diff
@@ -40,7 +40,7 @@ private bool CheckUserCredentials(User user, UserEmailPassword credentials)
             this.busControl = busControl;
         }
 
-        public async Task<string> Execute(UserEmailPassword data)
+        public async Task<UserToken> Execute(UserEmailPassword data)
         {
             User user = await GetUserFromUserService(data);
```
dTarasova: Why do we have to return user?

Services/AuthenticationService/Commands/LoginCommand.cs
```diff
@@ -49,7 +49,15 @@ public async Task<string> Execute(UserEmailPassword data)
                 throw new Exception("User not found =(");
```
dTarasova: Why don't custom exc?

Services/AuthenticationService/Controllers/AuthenticationController.cs
```diff
@@ -31,7 +33,9 @@ public string Login([FromServices] ILoginCommand command, [FromBody] UserEmailPa
         {
             var result = command.Execute(user);
 
-            return result.Result;
+            var userToken = result.Result;
+
+            return JsonSerializer.Serialize(userToken);
```
dTarasova: Why don't we check id here?  If we check it in dbs it is solid error. Why don't we get useremailpassword by email from database and then check if password hash the same here

dTarasova (changes requested):

Clients/GUI/Pages/SignIn.razor
nmamro: > Why github isn' satisfied?

About what?

Clients/GUI/Scripts/SignUpper.cs
nmamro: > I am not sure, but here we don't need try catch?

We try catch in the page's code

Clients/GUI/Shared/NavMenu.razor
nmamro: > Local Storage is different object for each user?

Yes it is, but that's to make sure that there's one user's info only
Idk, mb there's a better solution, then suggest changes

Services/AuthenticationService/Commands/LoginCommand.cs
nmamro: > Why do we have to return user?

To store id-token pair in client's storage

Services/AuthenticationService/Commands/LoginCommand.cs
zzzloba: > Why don't custom exc?

Because they suck right now, I updated them in another pr

Services/AuthenticationService/Controllers/AuthenticationController.cs
nmamro: > Why don't we check id here? If we check it in dbs it is solid error. Why don't we get useremailpassword by email from database and then check if password hash the same here

Why would we need to check id anywhere?
Client sends email and password hash and gets the id and token, which are correct because they come from DBS

Clients/GUI/Pages/SignIn.razor
dTarasova: Look at the files changed

Services/AuthenticationService/Controllers/AuthenticationController.cs
nmamro: > I suggest to change the name of result. What type it is? Why does result have result properties. It isn't understandable

It's Task with UserToken, what are suggestions?

Clients/GUI/Pages/SignUp.razor
nmamro: > Don't we suppose to catch customs exceptions? and have different handling for them?

Because our custom exceptions need to be updated (check new pr)

Services/AuthenticationService/Commands/LoginCommand.cs
nmamro: > Why don't custom exc?

Answered above

nmamro:
Make sure you can launch AuthenticationService, UserService, DatabaseService and GUI and after signing up in GUI check if there's a new user the database

Services/AuthenticationService/Commands/LoginCommand.cs
dTarasova: Let's think about that. Don't we have to add Id in UserEmailPassword and receive it?  I don't think that it is good idea to use user here

Services/AuthenticationService/Controllers/AuthenticationController.cs
nmamro: Fixing in next commit

Clients/GUI/Pages/SignIn.razor
nmamro: We don't really need  to wait until record is added to storage

Clients/GUI/Pages/SignUp.razor
nmamro: Fixing it in next commit

Services/AuthenticationService/Commands/LoginCommand.cs
nmamro: > Let's think about that. Don't we have to add Id in UserEmailPassword and receive it? I don't think that it is good idea to use user here

>  add Id in UserEmailPassword

You just described User class =)

Services/AuthenticationService/Commands/LoginCommand.cs
nmamro: Fixed

Clients/GUI/Pages/SignUp.razor
nmamro: Fixed

Services/AuthenticationService/Controllers/AuthenticationController.cs
nmamro: Fixed

Services/AuthenticationService/Controllers/AuthenticationController.cs
```diff
@@ -27,11 +30,11 @@ public AuthenticationController([FromServices] ITokensEngine tokenEngine)
         /// </summary>
         [Route("login")]
         [HttpPost]
-        public string Login([FromServices] ILoginCommand command, [FromBody] UserEmailPassword user)
+        public async Task<string> Login([FromServices] ILoginCommand command, [FromBody] UserEmailPassword user)
```
ifelseelif: I don't understand why it is async?

Clients/GUI/Pages/SignUp.razor
```diff
@@ -1,57 +1,74 @@
...
+            var userToken = SignInner.SignIn(credentials);
+            LocalStorage.SetItemAsync(userToken.UserId.ToString(), userToken.Body);
+            AuthenticationProvider.MarkSignedIn();
+            NavigationManager.NavigateTo("/userinfo");
+        }
+        catch (BadRequestException e)
+        {
+            userInput = new SignUpViewModel();
+            isErrorDisplayed = true;
+            errorMessage = "Something went wrong during registration";
```
ifelseelif, line 86: Maybe use a field message of exception instead of "Something went wrong during registration"?

Services/AuthenticationService/Controllers/AuthenticationController.cs
dTarasova: We have to make async all calls to not stop services work, as I understood

Clients/GUI/Pages/SignIn.razor
```diff
@@ -35,21 +32,23 @@
     bool isHidden = true;
     string errorMessage = "";
 
-    private void TrySignIn()
+    private async Task TrySignIn()
```
DanilFadeeev, line 34: why return's type is task and not void? Who use this task in main thread? Are you sure we need mark this method async? We can use getresult method instead await. I think, we musd  do it, because user can send only one pacage of data per time

DanilFadeeev (changes requested):

Clients/GUI/Pages/SignUp.razor
nmamro: > Maybe use a field message of exception instead of "Something went wrong during registration"?

I think user would like this one more than "403 Bad Request" or sth like that

Clients/GUI/Pages/SignIn.razor
nmamro: > why return's type is task and not void? Who use this task in main thread? Are you sure we need mark this method async? We can use getresult method instead await. I think, we musd do it, because user can send only one pacage of data per time

Check updated version of SignInner

Services/AuthenticationService/Controllers/AuthenticationController.cs
nmamro: > I don't understand why it is async?

Because there's asynchronous backend logic

Cerberous007 (approved):

nmamro:
Make sure that sign in and sign up work for you. To do this, do not forget to delete old TradingStation database before launching this updated version.

Clients/GUI/Pages/SignUp.razor
dTarasova: I think right now it is better to know where we have an error without debug. This message can be change later

nmamro: > I think right now it is better to know where we have an error without debug. This message can be change later

Why don't you like debug? You should print too much information in GUI anyway or you won't get enough information without the debug to find the exact place where error occurs

Clients/GUI/Pages/SignIn.razor
DanilFadeeev: i checked. What should I see?

DanilFadeeev (approved):

VLiamin (approved):

ifelseelif (approved):

dTarasova (approved):

PR #58, opened 2020-04-05
DataBaseService_FindUserById

Cerberous007:


Services/DataBaseService/BrokerConsumers/FindConsumer.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class FindConsumer : IConsumer<InternalFindUserRequest>
+    {
+        private readonly IUserRepository userRepository;
+
+        private UserCredential GetUserCredentialById(Guid Id)
+        {
+            return userRepository.GetUserCredentialById(Id);
+        }
+
+        public FindConsumer([FromServices] IUserRepository repository)
```
strannick-rodion, line 22: Why is the variable name not a userRepository?

Services/DataBaseService/BrokerConsumers/LoginConsumer.cs
```diff
@@ -10,21 +10,21 @@
...
         private UserCredential GetUserCredential(string email)
         {
-            return _repository.GetUserCredential(email);
+            return userRepository.GetUserCredential(email);
         }
 
         public LoginConsumer([FromServices] IUserRepository repository)
         {
-            _repository = repository;
+            userRepository = repository;
```
strannick-rodion: The same question

Common/DTO/BrokerRequests/InternalLoginRequest.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace DTO.RestRequests
+{
+    public class InternalLoginRequest
```
dTarasova, line 3: Why don;t use here UserCredentials

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace DataBaseService.BrokerConsumers
+{
+    public class FindConsumer : IConsumer<InternalFindUserRequest>
+    {
+        private readonly IUserRepository userRepository;
+
+        private UserCredential GetUserCredentialById(Guid Id)
```
dTarasova: Is this id of user tables or user credential? As I understood, client send us user id

dTarasova: Why private methods before public?

Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -26,13 +29,39 @@ public void CreateUser(User user)
...
+            }
+            
         }
 
         public UserCredential GetUserCredential(string email)
         {
             var dbCredential = dbContext.UsersCredentials.FirstOrDefault(uc => uc.Email == email);
+            if (dbCredential is null)
+            {
+                throw new ForbiddenException("User not found");
```
dTarasova: Is this is forbidden or bad request? It is user can't login, right?

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
DarkFelas: For what? Are there any other repositories?

Common/Kernel/Middlewares/CheckTokenMiddleware.cs
```diff
@@ -25,7 +25,7 @@ public CheckTokenMiddleware(RequestDelegate next, [FromServices] IBus bus)
         public async Task InvokeAsync(HttpContext context)
         {
             if (string.Equals(context.Request.Method, RestRequestType.POST.ToString(), StringComparison.OrdinalIgnoreCase)
-                && string.Equals(context.Request.Path, "/users/create", StringComparison.OrdinalIgnoreCase))
+                && string.Equals(context.Request.Path, "/user/create", StringComparison.OrdinalIgnoreCase))
```
DarkFelas: Revert this back, controller should have User**s**Controller name

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
DarkFelas: > Why private methods before public?

Because it is a good practice

DarkFelas: > Is this id of user tables or user credential? As I understood, client send us user id

Good question

Services/DataBaseService/Repositories/UserRepository.cs
DarkFelas: It is NotFound

Common/Kernel/Middlewares/CheckTokenMiddleware.cs
Cerberous007: It was an accident
Now fixed

Services/DataBaseService/Repositories/UserRepository.cs
Cerberous007: In the beginning, I put NotFound here, but I thought, that NotFound exception is issued in cases when the page is not found, and not some data not found.
But if NotFound is used correctly here, I can replace it)

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
Cerberous007: In constructor we explicitly specify the type of object. 
I named the field so as not confuse what type of repository it is

Services/DataBaseService/BrokerConsumers/LoginConsumer.cs
Cerberous007: See previous

Common/DTO/BrokerRequests/InternalLoginRequest.cs
Cerberous007: Because UserCredential contains an Id and userId, but initially they are not on client.

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
Cerberous007: I think, that client was sending us the UserCredential Id. We should check this with those who write the client.

strannick-rodion (approved):

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
dTarasova: ]But we have decided to write public then private as I remember

nmamro: > ]But we have decided to write public then private as I remember

We have this one indeed, mb we should consider changing it? I also don't think that private methods should go after public as it's more logical to write private first, because IMHO they are expected to be used in public methods so when you open a file and review the class, it's good to see all needed implementations before getting to usage of them in public methods

Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -22,17 +25,44 @@ public void CreateUser(User user)
...
+            else
+            {
+                throw new ForbiddenException("User already exists");
+            }
+            
         }
 
         public UserCredential GetUserCredential(string email)
         {
             var dbCredential = dbContext.UsersCredentials.FirstOrDefault(uc => uc.Email == email);
```
nmamro, line 54: May be we may ignore capital letters comparing emails?

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
DarkFelas: > I think, that client was sending us the UserCredential Id. We should check this with those who write the client.

All this about argument name, change it on userId or userCredentialId

Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -22,17 +25,44 @@ public void CreateUser(User user)
...
+            var dbCredential = dbContext.UsersCredentials.FirstOrDefault(uc => uc.Email == userCredential.Email);
+
+            if (dbCredential is null)
+            {
+                dbContext.UsersCredentials.Add(mapper.MapToDbUserCredential(userCredential));
+                dbContext.SaveChanges();
+            }
+            else
+            {
+                throw new ForbiddenException("User already exists");
```
DarkFelas: It is not a Forbidden type, it is BadRequest

Services/DataBaseService/BrokerConsumers/FindConsumer.cs
Cerberous007: Changed

Services/DataBaseService/Repositories/UserRepository.cs
Cerberous007: done

Services/DataBaseService/Repositories/UserRepository.cs
Cerberous007: done

ifelseelif (approved):
seems ok

Common/DTO/BrokerRequests/InternalFindUserRequest.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DTO.BrokerRequests
+{
+    public class InternalFindUserRequest
+    {
+        public Guid Id { get; set; }
```
dTarasova, line 9: How about change name here also?

Common/DTO/RestRequests/FindUserRequest.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DTO.RestRequests
+{
+    public class FindUserRequest
+    {
+        public Guid Id { get; set; }
+    }
```
dTarasova, line 10: change name

Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -22,17 +25,46 @@ public void CreateUser(User user)
...
             var dbCredential = dbContext.UsersCredentials.FirstOrDefault(uc => uc.Email == email);
+            if (dbCredential is null)
+            {
+                throw new NotFoundException("User not found");               
+            }
+
+            return mapper.MapUserCredential(dbCredential);
+        }
+
+        public UserCredential GetUserCredentialById(Guid userCredentialId)
```
nmamro, line 63: I want to be able to get not only credentials, but also name, surname etc by id

nmamro (changes requested):

PR #62, opened 2020-04-05
DataBaseService_DeleteUserLogic_ChangeToUserId

strannick-rodion:


nmamro (changes requested):
After merge of previous request without me, I gonna request changes here so I can check it properly tomorrow so no one can merge it

dTarasova (approved):

nmamro (changes requested):
It does not work, fix and don't try to push bugs to develop

zzzloba:
Why after new user is created, their IsActive = false?

nmamro:
Change logic of Delete in Repository, because UserId is not a primary key in UsersCredentials, so we can not just call `.Find(userId)` to find a record there

strannick-rodion:
This work

PR #64, opened 2020-04-05
UserService_DeleteUserMassTransitAdding

dTarasova:


Cerberous007 (approved):
Maybe I missed something, but ovaerall is ok.

nmamro (commented):
Seems fine overall, have you tested it?

Services/UserService/Validators/DeleteUserRequestValidator.cs
```diff
@@ -0,0 +1,17 @@
...
+using FluentValidation;
+
+namespace UserService.Validators
+{
+    public class DeleteUserRequestValidator: AbstractValidator<DeleteUserRequest>
+    {
+        public DeleteUserRequestValidator()
+        {
+            RuleFor(request => request.UserId)
+                .NotNull()
```
ifelseelif, line 11: NotEmpty ensures that the object isn't null.

ifelseelif (approved):

nmamro (approved):

Services/UserService/Validators/DeleteUserRequestValidator.cs
dTarasova: > Seems fine overall, have you tested it?
Yeah, it's okay

VLiamin (approved):

PR #65, opened 2020-04-06
DataBaseService_GetUserById

Cerberous007:


Common/DTO/BrokerRequests/InternalGetUserByIdRequest.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace DTO.BrokerRequests
+{
+    public class InternalGetUserByIdRequest
```
dTarasova, line 5: Maybe we should combine this and InternalDeleteUserRequest. They are absolutely the same

Services/DataBaseService/Mappers/UserMapper.cs
```diff
@@ -31,7 +31,13 @@ public DbUserCredential MapToDbUserCredential(UserCredential userCredential)
 
         public User MapUser(DbUser dbUser)
         {
-            throw new NotImplementedException();
+            return new User
```
dTarasova, line 34: User doesn't have email here?

Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -28,19 +28,44 @@ public void CreateUser(User user)
...
+            return mapper.MapUser(dbUser);
+        }
+
+        public void DeleteUser(Guid userId)
         {
-            var dbUserCredential = dbContext.Find<DbUserCredential>(userIdCredential);
+            var dbUserCredential = dbContext.Find<DbUserCredential>(userId);
             if (dbUserCredential == null)
             {
                 throw new ForbiddenException("Not found User for delete");
```
dTarasova: Forbidden exception with message Not found is a little strange

Common/DTO/BrokerRequests/InternalGetUserByIdRequest.cs
Cerberous007: I am thinking about this and want to discuss it with our colleagues.

Services/DataBaseService/Mappers/UserMapper.cs
Cerberous007: Currently dbUser does not have an email address and we discussed the option to save email on the client.
If necessary, we can add it here, but it will be a little more difficult

Services/DataBaseService/Repositories/UserRepository.cs
Cerberous007: It is not my part of code

Cerberous007: But done

Common/DTO/BrokerRequests/InternalGetUserByIdRequest.cs
nmamro: You can decide either to create different endpoints or to use different models. I actually like different models more now, may be in perspective my opinion is wrong xD

Services/DataBaseService/BrokerConsumers/CreateUserConsumer.cs
```diff
@@ -19,7 +19,7 @@ public CreateUserConsumer([FromServices] IUserRepository userRepository)
         private OperationResult CreateUser(InternalCreateUserRequest request)
         {
             userRepository.CreateUser(request.User);
-            userRepository.CreateUserCredential(request.Credential);
+            userRepository.CreateUserCredential(request.Credential);
```
nmamro, line 22: What is this space? 
Install this [extension](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer), it's cool

Services/DataBaseService/BrokerConsumers/DeleteUserConsumer.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace DataBaseService.BrokerConsumers
 {
-    public class DeleteUserConsumer : IConsumer<InternalDeleteUserRequest>
+    public class DeleteUserConsumer : IConsumer<InternalGetUserByIdRequest>
```
nmamro: So you decided to use same internal request model for get and delete.
I thought that if we have same request model, we decided to create different endpoints and then check endpoint in one consumer, which handles the request.
So seeing this now, I'm not even sure that it will work properly, because I can't see different endpoints configuration in DBS (but mb I'm wrong, then correct me)

Services/DataBaseService/Mappers/UserMapper.cs
dTarasova: Why we should save it on client?

Services/DataBaseService/Startup.cs
```diff
@@ -53,6 +53,7 @@ private IBusControl CreateBus(IServiceProvider serviceProvider)
                     ep.ConfigureConsumer<CreateUserConsumer>(serviceProvider);
                     ep.ConfigureConsumer<DeleteUserConsumer>(serviceProvider);
                     ep.ConfigureConsumer<LoginUserConsumer>(serviceProvider);
+                    ep.ConfigureConsumer<GetUserByIdConsumer>(serviceProvider);
```
nmamro, line 56: Yeah, look here. You have 2 consumers here in the same endpoint, receiving same request model, will this code work at all? Have you tested it?

nmamro (changes requested):

Common/DTO/BrokerRequests/InternalGetUserByIdRequest.cs
Cerberous007: I also like option with different models

Services/DataBaseService/BrokerConsumers/CreateUserConsumer.cs
Cerberous007: Thanks for link)

Services/DataBaseService/Mappers/UserMapper.cs
Cerberous007: Because this information can be necessary for another actions.
I'll check with management)

Services/DataBaseService/BrokerConsumers/DeleteUserConsumer.cs
Cerberous007: I don't really know what you mean, but I use different models

Services/DataBaseService/Startup.cs
Cerberous007: Fixed it

Services/DataBaseService/BrokerConsumers/DeleteUserConsumer.cs
Cerberous007: Fixed

Services/DataBaseService/Mappers/UserMapper.cs
Cerberous007: This issue has already been resolved

LK777LK (approved):

dTarasova (approved):

nmamro (approved):
Seems ok for me

nmamro:
But update from dev =)

Services/DataBaseService/BrokerConsumers/CreateUserConsumer.cs
denis-kriv, line 22: Why do we need this space

Maksim-Trolina (approved):

PR #67, opened 2020-04-06
DataBaseService_DeleteUserLogicWithMassTrasnit_BugFixes

strannick-rodion:


Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -38,16 +38,16 @@ public UserCredential GetUserCredential(string email)
 
             return mapper.MapUserCredential(dbCredential);
         }
-        public void DeleteUser(Guid userIdCredential)
+        public void DeleteUser(Guid userId)
```
nmamro, line 41: It would look prettier if there was an empty line between methods

Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -38,16 +38,16 @@ public UserCredential GetUserCredential(string email)
...
-            var dbUserCredential = dbContext.Find<DbUserCredential>(userIdCredential);
+            var dbUserCredential = dbContext.UsersCredentials.FirstOrDefault(uc => uc.UserId == userId);
             if (dbUserCredential == null)
             {
                 throw new ForbiddenException("Not found User for delete");
             }
             if(dbUserCredential.IsActive==false)
             {
-                throw new ForbiddenException("User was deleted early");
+                throw new ForbiddenException("User was deleted early or not confirmed");
```
nmamro, line 50: I still don't really understand "or not comfirmed part" to be honest

nmamro (approved):
Good =)

Services/DataBaseService/Repositories/UserRepository.cs
Cerberous007: In order for the user to have an IsActive field other than 0, they must confirm themselves using email.
And in this case, If IsActive is 0, it`s mean that user was deleted early or was not confirmed)

Cerberous007 (approved):

Services/DataBaseService/Repositories/UserRepository.cs
nmamro: > In order for the user to have an IsActive field other than 0, they must confirm themselves using email.
> And in this case, If IsActive is 0, it`s mean that user was deleted early or was not confirmed)

Ah, I got it know, thanks =)

dTarasova (approved):

denis-kriv (approved):

Maksim-Trolina (approved):

PR #68, opened 2020-04-06
UserService_AddTestsToValidator

DanilFadeeev:


Cerberous007 (commented):
I'm very glad that you made a new branch that doesn't have extra commits.
But the name is still not good(

Services/UserService/Controllers/UsersController.cs
```diff
@@ -19,6 +21,16 @@ public class UsersController : ControllerBase
         [HttpPost]
         public async Task<bool> CreateUser([FromServices] ICreateUserCommand command, [FromBody] CreateUserRequest request)
```
dTarasova, line 22: Please don't write logic in the controller, it is already done in CreateUserCommand

Tests/AuthServiceTests/Validators/CreateUserRequestValidator.cs
```diff
@@ -0,0 +1,130 @@
+﻿using DTO.RestRequests;
+using NUnit.Framework;
+using System;
+using UserService.Validators;
+using static UserService.Validators.ErrorsMessages;
+
+
+namespace AuthenticationServiceTests.Validators
```
dTarasova, line 8: Why this namespace

dTarasova: And in Auth service

Tests/AuthServiceTests/Validators/CreateUserRequestValidator.cs
```diff
@@ -0,0 +1,130 @@
...
+{
+    class CreateUserRequestValidatorTests
```
dTarasova, line 10: I think we already have the same validator , but it named incorrectly: emailPasswordValidatorTests

dTarasova (changes requested):

Tests/AuthServiceTests/Validators/CreateUserRequestValidator.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        CreateUserRequestValidator validator;
```
nmamro, line 12: Public/private/iternal?

Services/UserService/Controllers/UsersController.cs
```diff
@@ -19,6 +21,16 @@ public class UsersController : ControllerBase
         [HttpPost]
         public async Task<bool> CreateUser([FromServices] ICreateUserCommand command, [FromBody] CreateUserRequest request)
         {
+            CreateUserRequestValidator validator = new CreateUserRequestValidator();
```
nmamro, line 24: Inject validator, then add from services

Services/UserService/Controllers/UsersController.cs
```diff
@@ -19,6 +21,16 @@ public class UsersController : ControllerBase
         [HttpPost]
         public async Task<bool> CreateUser([FromServices] ICreateUserCommand command, [FromBody] CreateUserRequest request)
         {
+            CreateUserRequestValidator validator = new CreateUserRequestValidator();
+            var validationResult = await validator.ValidateAsync(request);
```
nmamro, line 25: There's Validate and throw for these occasions, so mb use it?

Tests/AuthServiceTests/Validators/CreateUserRequestValidator.cs
```diff
@@ -0,0 +1,130 @@
...
+        CreateUserRequest correctData;
+        CreateUserRequest incorrectBirthday;
+        CreateUserRequest incorrectEmail;
+        CreateUserRequest incorrectFirstName;
+        CreateUserRequest incorrectLastName;
+        CreateUserRequest emptyPassword;
+
+        [SetUp]
```
nmamro, line 20: This is called before each test

nmamro (changes requested):

Services/UserService/Controllers/UsersController.cs
DanilFadeeev: why?

PR #70, opened 2020-04-07
GUI_UserService_Get user integration

nmamro:


Clients/GUI/Pages/SignUp.razor
```diff
@@ -102,5 +102,7 @@
     protected override void OnInitialized()
     {
         userInput = new SignUpViewModel();
+
+        userInput.Birthday = DateTime.Today;
```
dTarasova: Do we have to write it here? How about write default value in constructor

Common/Kernel/RestClient.cs
```diff
@@ -32,7 +32,7 @@ statusCode switch
 
         private async Task<TOut> GetResponse()
```
dTarasova: If you start to fix this class, can you change variables name to codeConvention

Services/UserService/Commands/DeleteUserCommand.cs
```diff
@@ -28,16 +28,16 @@ public async Task<bool> Execute(DeleteUserRequest request)
...
             {
                 throw new BadRequestException("Unable to delete user.");
             }
             return deleteUserResult;
         }
 
-         
-         private async Task<bool> deleteUser(InternalDeleteUserRequest request)
+
+         private async Task<bool> DeleteUser(InternalDeleteUserRequest request)
```
dTarasova, line 40: it is private method , why do you change the name?

Services/UserService/Controllers/UsersController.cs
```diff
@@ -35,9 +35,9 @@ public async Task<bool> DeleteUser([FromServices] IDeleteUserCommand command, [F
 
         [Route("get")]
         [HttpGet]
-        public User GetUser([FromHeader] Guid userId)
+        public async Task<User> GetUser([FromServices] IGetUserByIdCommand command, [FromQuery] Guid userId)
```
dTarasova: Why do you use  Guid here and not getUserByIdRequest in RestRequests? 

And why from query and not header?

nmamro: > Why do you use Guid here and not getUserByIdRequest in RestRequests?
> 
> And why from query and not header?

Fixed, my bad

Services/UserService/Commands/DeleteUserCommand.cs
nmamro: > it is private method , why do you change the name?

Because it's method?

Clients/GUI/Pages/SignUp.razor
nmamro: > Do we have to write it here? How about write default value in constructor

Fixed

Common/Kernel/RestClient.cs
nmamro: > If you start to fix this class, can you change variables name to codeConvention

Removed changes

Services/UserService/Commands/DeleteUserCommand.cs
dTarasova: So? I thought that was a common rule for all part of code

nmamro: > So? I thought that was a common rule for all part of code

Well, my bad, i might have missed that in code convention. But i don't like this part, i will suggest to discuss this

dTarasova (approved):

Maksim-Trolina (approved):

Clients/GUI/Authentication/AuthStateProvider.cs
```diff
@@ -10,20 +11,21 @@ namespace GUI.Authentication
     /// </summary>
     public class AuthStateProvider : AuthenticationStateProvider
     {
-        private readonly ILocalStorageService localStorage;
```
ifelseelif, line 13: Why are you changing to this?

nmamro: > Why are you changing to this?

Because I decided that right now we don't want data on client side to be able to live that long as local storage makes it possible, so it's kind of overkill. If we will need local storage over session storage later, it can be changed in 1 minute

ifelseelif (approved):

LK777LK (approved):

PR #71, opened 2020-04-07
DbService CreateUserTests

LK777LK:


DanilFadeeev (approved):

Services/DataBaseService/Database/Models/DbUser.cs
```diff
@@ -1,16 +1,70 @@
...
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace DataBaseService.Database.Models
 {
-    public class DbUser
+    public class DbUser : IEquatable<DbUser>
     {
+        public static bool operator ==(DbUser first, DbUser second)
```
nmamro: Why you updated the class (and not only this one), will anyone need comparison anywhere apart from tests?

Common/DTO/User.cs
```diff
@@ -2,12 +2,51 @@
 
 namespace DTO
 {
-    public class User
+    public class User : IEquatable<User>
```
nmamro: Don't need this interface

Common/DTO/User.cs
```diff
@@ -2,12 +2,51 @@
...
     {
         public Guid Id { get; set; }
         public string Email { get; set; }
         public string FirstName { get; set; }
         public string LastName { get; set; }
         public DateTime Birthday { get; set; }
+
+        public bool Equals(User other)
```
nmamro: I think it's better to make checker in tests class, remove this code from data transfer objects please

nmamro (changes requested):

Common/DTO/User.cs
nmamro: Then you will have `AreUsersEqual()` in tests and you will call `Assert.IsTrue()` for example

Tests/DatabaseServiceTests/Mappers/UserMapperTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        IUserMapper mapper;
+
+        User user;
+        DbUser dbUser;
+
+        UserCredential credential;
+        DbUserCredential dbCredential;
+
+        [SetUp]
```
ifelseelif: All these objects will be initialized every time. Do you need it ?

Services/DataBaseService/Database/Models/DbUser.cs
```diff
@@ -1,16 +1,70 @@
 ﻿using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace DataBaseService.Database.Models
 {
-    public class DbUser
+    public class DbUser : IEquatable<DbUser>
```
ifelseelif: I do not understand why you need it ?

VLiamin (approved):

ifelseelif (approved):

PR #76, opened 2020-04-07
DataBaseService_CreateUser_EmailUniqueCheck

dTarasova:


Maksim-Trolina (approved):

strannick-rodion (approved):

denis-kriv (approved):

Services/DataBaseService/Repositories/UserRepository.cs
```diff
@@ -19,8 +19,10 @@ public UserRepository(IUserMapper mapper, TPlatformDbContext dbContext)
             this.dbContext = dbContext;
         }
 
-        public void CreateUser(User user)
+        public void CreateUser(User user, string email)
         {
+            if (dbContext.UsersCredentials.Any(userCredential => userCredential.Email == email))
```
Cerberous007, line 24: I think we should do a check that ignores capital letters

Cerberous007 (changes requested):

Services/DataBaseService/Repositories/UserRepository.cs
dTarasova: Why?

dTarasova: Isn't email Hey@there.com and hey@there.com are different?

Cerberous007: No, it`s same emails. And you should avoid it.

Cerberous007: In our db this is different emails, but not in other world))

dTarasova: Hmmm, Okay, I checked in some places and you are right. Thanks :)

dTarasova: This request is translated to sql, and sql doesn't compare case of letter. So that would be checked automatically

Cerberous007: Hmmm, okey

Cerberous007 (approved):

PR #78, opened 2020-04-08
BrokerService_GetCurrency

denis-kriv:


Integrations/CentralBankIntegrationLib/CentralBankNewsPublisher.cs
```diff
@@ -40,29 +40,29 @@ public string GetNews()
...
+        {
+            var url = "http://www.cbr.ru/scripts/XML_daily.asp";
+
+            HttpWebRequest httpWebRequest = (HttpWebRequest)WebRequest.Create(url);
+            HttpWebResponse httpWebResponse = (HttpWebResponse)httpWebRequest.GetResponse();
+
+            using Stream responseStream = httpWebResponse.GetResponseStream();
+            using StreamReader streamReader = new StreamReader(responseStream);
+
+            var top5Valutes = new List<Valute>();
```
Cerberous007, line 59: The numbers in the name look strange

Services/BrokerService/Constants.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace BrokerService
+{
+    public static class Constants
+    {
+        internal static int Depth = 5;
+        public static string Token = "example ";
```
Cerberous007: Why do you need this class? Why constant token is "example"?

denis-kriv: If the token changes, we can easily change it in one place, the same with the depth of the glass

denis-kriv: We will change the token when we recognize it

Integrations/CentralBankIntegrationLib/CentralBankNewsPublisher.cs
denis-kriv: I did not change this file!!)

Services/BrokerService/Constants.cs
nmamro: > We will change the token when we recognize it

Why not to use appsettings?

Services/BrokerService/BrokerService.csproj.BASE
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
nmamro: Use git properly

Integrations/CentralBankIntegrationLib/CentralBankNewsPublisher.cs
nmamro: > I did not change this file!!)

Then you better find out why it is here

Services/BrokerService/Controllers/BrokerController.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Collections.Generic;
+using BrokerServices;
+using BrokerServices.Utils;
+using Interfaces;
+using Microsoft.AspNetCore.Mvc;
+
+namespace BrokerService.Controllers
+{
+    [Route("api/[controller]")]
```
ifelseelif: why do you use api ?

nmamro:
Guys, fix this request please, it's becoming too old

Services/BrokerService/BrokerService.csproj.BASE
denis-kriv: fixed

Services/BrokerService/Constants.cs
denis-kriv: fixed

Services/BrokerService/Controllers/BrokerController.cs
denis-kriv: fixed

PR #80, opened 2020-04-09
NewsService_Added get news logic and refactored get currencies a bit

nmamro:


Services/NewsService/Utils/NewsPublisherFactory.cs
```diff
@@ -15,8 +15,8 @@ public static INewsPublisher Create(NewsPublisherTypes newsPublisherType)
         {
             switch (newsPublisherType)
             {
-                case NewsPublisherTypes.CentralBank:
-                    return new RussianCBInfo();
+                case NewsPublisherTypes.Rambler:
+                    return new RamblerRssReader();
                 default:
                     //TODO change on custom exception
```
ifelseelif: The time has come. Change to custom exception

nmamro: > The time has come. Change to custom exception

Changed

Common/DTO/NewsItem.cs
```diff
@@ -0,0 +1,17 @@
...
+
+        public string PubDate { get; set; }
+
+        public string Description { get; set; }
+
+        public string Category { get; set; }
+
+        public string Author { get; set; }
+    }
+}
```
LK777LK, line 17: I would add attributes [JsonPropertyName(...)] here. Who can explain it well or not?

Common/Interfaces/INewsPublisher.cs
```diff
@@ -9,6 +8,6 @@ namespace Interfaces
     /// </summary>
     public interface INewsPublisher
     {
-        List<ExchangeRate> GetCurrencies();
+        IEnumerable<NewsItem> GetNews(string feedUrl);
     }
 }
```
LK777LK, line 13: Why you use IEnumerable<T> and not specific type?

Services/NewsService/Utils/NewsPublisherFactory.cs
LK777LK: You can use new way switches write

LK777LK (approved):

Common/Interfaces/INewsPublisher.cs
nmamro: > Why you use IEnumerable and not specific type?

Hm, I made it IEnumerable because I did not see any reason to return more complicated class such as List here, but I might be wrong =)

Common/DTO/NewsItem.cs
nmamro: > I would add attributes [JsonPropertyName(...)] here. Who can explain it well or not?

Isn't it Newtonsoft.Json thing?

Services/NewsService/Utils/NewsPublisherFactory.cs
nmamro: > You can use new way switches write

Updated 😃

Common/DTO/NewsItem.cs
LK777LK: no, it is system.text.json.serialize

nmamro: > no, it is system.text.json.serialize

As I understood, It's a way to explicitly say how you want properties to be named in json object, and by default they should be names just as you write in c#

Cerberous007 (approved):

ifelseelif (approved):

Maksim-Trolina (approved):

PR #90, opened 2020-04-11
Kernel_DataBaseService_AddLoggerLogic

Cerberous007:


Common/Kernel/LoggingEngine/LogMessage.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+
+namespace Kernel
+{
+    public class LogMessage
+    {
+        public Guid Id { get; set; }
+        public Guid? ParentId { get; set; }
```
ifelseelif, line 9: Why do we need ParentId?

Common/Kernel/LoggingEngine/Logger.cs
```diff
@@ -0,0 +1,56 @@
...
+        public Logger(IServiceProvider provider)
+        {
+            this.provider = provider;
+        }
+
+        private async void AddLog(LogMessage log)
+        {
+            var bus = provider.GetRequiredService<IBus>();
+
+            var uri = new Uri("rabbitmq://localhost/DatabaseService_Logs");
```
dTarasova, line 22: Do we have to log only dbservice?

Common/Kernel/LoggingEngine/Logger.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            var endpont = await bus.GetSendEndpoint(uri);
+            await endpont.Send(log);
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
```
dTarasova, line 28: What is this method do?

Common/Kernel/LoggingEngine/LoggerProvider.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+
+namespace Kernel.LoggingEngine
+{
+    public class LoggerProvider: ILoggerProvider
+    {
+        private readonly IServiceProvider _provider;
```
dTarasova: How about change var name accordin to code convention

Services/AuthenticationService/Controllers/AuthenticationController.cs
```diff
@@ -21,6 +29,8 @@ public class AuthenticationController : ControllerBase
         [HttpPost]
         public async Task<UserToken> Login([FromServices] ILoginCommand command, [FromBody] LoginRequest request)
         {
+            _logger.LogInformation("Success");
```
dTarasova: Why do we right success before command will be execute?

Services/AuthenticationService/Controllers/AuthenticationController.cs
```diff
@@ -6,13 +6,21 @@
 using DTO;
 using DTO.RestRequests;
 using AuthenticationService.Interfaces;
+using Microsoft.Extensions.Logging;
 
 namespace AuthenticationService.Controllers
```
dTarasova, line 11: You add this stuff only in Auth controller , who suppose to add this stuff in user service?

Common/Kernel/LoggingEngine/LogMessage.cs
Cerberous007: To track call chains if necessary

Common/Kernel/LoggingEngine/Logger.cs
Cerberous007: This stuff need for sending logs for writing to the database from other services

Services/AuthenticationService/Controllers/AuthenticationController.cs
Cerberous007: This is just an example that Spartak wrote. My task was only to create a LoggerEngine and log entry logic to the database. We will need to enable logging on our own

Common/Kernel/LoggingEngine/Logger.cs
Cerberous007: Nothing, because I don`t use it)) I need to implement the ILogger interface methods

Services/AuthenticationService/Controllers/AuthenticationController.cs
Cerberous007: It`s just example. This means that we were able to reach this method

Common/Kernel/LoggingEngine/LoggerProvider.cs
Cerberous007: Fixed

dTarasova (approved):

strannick-rodion (approved):

ifelseelif (approved):

LK777LK (approved):

PR #101, opened 2020-04-17
Gui_SignInTests

ifelseelif:


SingInTests/Program.cs
```diff
@@ -0,0 +1,71 @@
...
+    internal class SignInTest
+    {
+        private CreateUserRequest request = new CreateUserRequest
+        {
+            FirstName = "Test",
+            LastName = "Test",
+            Birthday = DateTime.Today.AddYears(-20),
+            Email = "testUser@mail.ru",
+            Password = "password"
+        };
```
dTarasova, line 22: I think better to add empty line here

SingInTests/Program.cs
```diff
@@ -0,0 +1,71 @@
...
+        private static void Main(string[] args)
+        {
+            var test = new SignInTest();
+            try
+            {
+                test.TryToSingUp();
+            }
+            catch (Exception e)
```
dTarasova, line 31: Do we have different types of exceptions? Why don't custom ?

SingInTests/Program.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                Console.WriteLine("Successful");
+            }
+            else
+            {
+                Console.WriteLine("Something was wrong");
+            }
+        }
+
+        private async Task TryToSingUp()
```
dTarasova, line 46: Pull request named sign in tests. But you write logic for sign up. Something went wrong :)

ifelseelif: How can I log in if the user does not exist in the database?

SingInTests/Program.cs
ifelseelif: Why do we need custom exception ?

SingInTests/Program.cs
ifelseelif: fixed

SingInTests/Program.cs
dTarasova: It is logical :)

SingInTests/Program.cs
dTarasova: I wasn't sure, try to understand how gui tests works

dTarasova (approved):

SingInTests/Program.cs
ifelseelif: This test simulates entering the necessary data into the login and password and then clicking send

SingInTests/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+            const string url = "https://localhost:5011/users/create";
+
+            var client = new RestClient<CreateUserRequest, bool>(url, RestRequestType.POST);
+
+            await client.Execute(request);
+        }
+
+        private bool TryToSingIn()
+        {
+            IWebDriver driver = new ChromeDriver();
```
denis-kriv, line 58: why not use var?

denis-kriv (approved):

Maksim-Trolina (approved):

SingInTests/Program.cs
ifelseelif: I didn't use var, because I need this interface. This interface allows me to quickly replace the driver

SingInTests/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        private async Task TryToSingUp()
+        {
+
+            const string url = "https://localhost:5011/users/create";
+
+            var client = new RestClient<CreateUserRequest, bool>(url, RestRequestType.POST);
+
+            await client.Execute(request);
```
Cerberous007, line 53: Mb I dont understand something, but is it normal to create new user in our system in each test? When we run this test, we will add test user to database, Correct me If I`m wrong

Cerberous007 (changes requested):

SingInTests/Program.cs
ifelseelif: Every time you run a test, it creates the same user.

Cerberous007: Okey, probably it`s not bad for sign in tests

Cerberous007 (approved):

PR #102, opened 2020-04-17
AuthService_PostmanTests

denis-kriv:


DanilFadeeev (approved):

Maksim-Trolina (approved):

ifelseelif (changes requested):
I think you should  make 1 more request, where do you create user with this email and password, because without it tests fall down

ifelseelif (approved):

LK777LK (approved):

PR #103, opened 2020-04-18
GUI_InstrumentsPage

nmamro:


Clients/GUI/Pages/Instruments.razor
```diff
@@ -0,0 +1,142 @@
...
+        else
+        {
+            <h3 class="m-lg-5 badge-info">Can't load instruments, try again later</h3>
+        }
+    </div>
+</div>
+
+
+@code {
+    public IEnumerable<Instrument> MarketInstruments { get; set; }
```
dTarasova: Do we need public here and in other fields?

Common/DTO/BankBrokerObjects/Instrument.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace DTO
+{
+    /// <summary>
+    /// Represents market instrument
+    /// </summary>
+    public class Instrument
+    {
+        public virtual string Figi { get; set; }
+
+        public virtual string Ticker { get; set; }
```
dTarasova: What is this mean?

Common/DTO/BankBrokerObjects/Instrument.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public virtual string Isin { get; set; }
```
dTarasova: And this

Common/DTO/BankBrokerObjects/Instrument.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public virtual string Type { get; set; }
+
+        public virtual string Name { get; set; }
+
+        public virtual string Currency { get; set; }
+
+        public virtual int Lot { get; set; }
```
dTarasova: And this :)

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
```diff
@@ -1,120 +1,77 @@
...
+                var conn = ConnectionFactory.GetSandboxConnection(token);
+                context = conn.Context;
 
-                instruments.Add(new TinkoffInstrumentAdapter(instrumentsList.Instruments[i],
-                    orderbook));
+                Depth = depth;
+            }
+            catch (Exception)
+            {
+                throw new BadRequestException();
```
dTarasova, line 27: Is this bad request exception? I think it is more like forbidden

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
```diff
@@ -1,120 +1,77 @@
...
-        }
+            MarketInstrumentList instrumentsList = tinkoffInstrumentType switch
+            {
+                InstrumentType.Bond => context.MarketBondsAsync().Result,
+                InstrumentType.Currency => context.MarketCurrenciesAsync().Result,
+                InstrumentType.Stock => context.MarketStocksAsync().Result,
+                _ => throw new BadRequestException()
+            };
+
+            Parallel.ForEach(instrumentsList.Instruments,
```
dTarasova, line 55: Why do we use parallel?

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
```diff
@@ -1,120 +1,77 @@
...
+                        instruments.Add(
+                            new TinkoffInstrumentAdapter(
+                                tinkoffInstrumentType,
+                                instrument,
+                                context.MarketOrderbookAsync(
+                                    instrument.Figi,
+                                    Depth).Result)
+                            );
+                    }
+                    catch (Exception) { }
```
dTarasova, line 66: why we don't throw here anything

Services/BrokerService/Commands/GetInstrumentsCommand.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using BrokerService.Utils;
+using Kernel.CustomExceptions;
+using DTO;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+
+namespace BrokerService.Commands
+{
+    public class GetInstrumentsCommand : IGetInstrumentsCommand
```
dTarasova: Can we use just ICommand interface?

Services/BrokerService/Commands/GetInstrumentsCommand.cs
```diff
@@ -0,0 +1,36 @@
...
+        public GetInstrumentsCommand([FromServices] ILogger<GetInstrumentsCommand> logger)
+        {
+            this.logger = logger;
+        }
+
+        public IEnumerable<Instrument> Execute(BankType bank, string token, int depth, string instrument)
+        {
+            try
+            {
+                logger.LogInformation($"Received request for getting {instrument} from {bank} with given depth {depth}.");
```
dTarasova: You always receive request  so you can log this info outside try catch block

Services/BrokerService/Controllers/BrokersController.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace BrokerService.Controllers
+{
+    [ApiController]
+    [Route("[controller]")]
+    public class BrokersController : ControllerBase
+    {
+        [Route("instruments/get")]
+        [HttpGet]
+        public IEnumerable<Instrument> GetInstruments([FromServices] IGetInstrumentsCommand command, [FromQuery] BankType bank, [FromQuery] string token, [FromQuery] int depth, [FromQuery] string instrument)
```
dTarasova: difficult to read

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
```diff
@@ -1,120 +1,77 @@
...
 using System.Threading.Tasks;
 using DTO;
 using Interfaces;
+using Kernel.CustomExceptions;
 using Tinkoff.Trading.OpenApi.Models;
 using Tinkoff.Trading.OpenApi.Network;
 
 namespace TinkoffIntegrationLib
 {
     public class TinkoffBankBroker : IBroker
```
dTarasova, line 12: Not sure, but do we need any logs here?

Clients/GUI/Scripts/BrokerManager.cs
```diff
@@ -0,0 +1,28 @@
...
+using Kernel;
+using Kernel.Enums;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace GUI.Scripts
+{
+    public static class BrokerManager
+    {
+        public static async Task<IEnumerable<Instrument>> GetInstruments(BankType bank, string token, string instrument, int depth = 10)
```
dTarasova: Difficult to read. Split into 2 lines

Clients/GUI/Scripts/BrokerManager.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            const string url = "https://localhost:5003/brokers/instruments/get";
+            //[FromQuery] BankType bank, [FromQuery] string token, [FromQuery] int depth, [FromQuery] InstrumentType instrument
+            var queryParams = new Dictionary<string, string>
```
dTarasova, line 20: Do we need logs in GUI?

Clients/GUI/Pages/Instruments.razor
```diff
@@ -0,0 +1,142 @@
...
+        AreInstrumentsLoading = true;
+
+        var result = await BrokerManager.GetInstruments(BankType.TinkoffBank, token, SelectedInstrument);
+
+        AreInstrumentsLoading = false;
+
+        MarketInstruments = result;
+    }
+
+    private void SaveInstrument(Instrument instrument)
```
dTarasova, line 128: What do you mean by Save Instrument to portfolio? Can you (now or in the future) buy instrument from this page? Or portfolio it is all instruments that you interested in?

dTarasova (changes requested):
Problems with unhandled exceptions

Services/BrokerService/Commands/GetInstrumentsCommand.cs
```diff
@@ -0,0 +1,36 @@
+﻿using BrokerService.Interfaces;
+using System.Collections.Generic;
+using System;
+using BrokerService.Utils;
+using Kernel.CustomExceptions;
+using DTO;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
```
dTarasova: Do you use  it?

nmamro: > Do you use it?

That's for adding logging

Clients/GUI/Pages/Instruments.razor
nmamro: > What do you mean by Save Instrument to portfolio? Can you (now or in the future) buy instrument from this page? Or portfolio it is all instruments that you interested in?

I had a task to be able to save instrument to portfolio, idk about further logic of saving

Clients/GUI/Scripts/BrokerManager.cs
nmamro: > Do we need logs in GUI?

My bad, forgot to remove it, fixing next commit

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
nmamro: > Not sure, but do we need any logs here?

Hm, actually I thought that we use logs in services, that's a library, I'll think about it and respond later

Common/DTO/BankBrokerObjects/Instrument.cs
nmamro: > And this :)

These names are not generated by my randomly, check, for example, tinkoff docs

Clients/GUI/Pages/Instruments.razor
nmamro: > Do we need public here and in other fields?

In most examples I've seen they use public stuff here, idk, seems like common practice

Common/DTO/BankBrokerObjects/Instrument.cs
nmamro: > What is this mean?

Answered lower

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
nmamro: > Is this bad request exception? I think it is more like forbidden

I guess you are right, will fix it soon

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
nmamro: > Why do we use parallel?

Because synchronously it will take too much time

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
nmamro: > why we don't throw here anything

Because for some reason Tinkoff API throws weird exceptions here, I have no idea why they do it
There's some instrument's parameter like MinPriceIncrement or sth like that and if it's 0, they throw some exception when we try to get the instrument

Services/BrokerService/Commands/GetInstrumentsCommand.cs
nmamro: > Can we use just ICommand interface?

Well why are you asking it here after approving lots of pull requests with different interfaces, am I missing sth?  I thought it's k, correct me if I'm wrong

Services/BrokerService/Commands/GetInstrumentsCommand.cs
nmamro: > You always receive request so you can log this info outside try catch block

What's the difference?

Services/BrokerService/Controllers/BrokersController.cs
nmamro: > difficult to read

I will split it in different lines, ok

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
nmamro: > Is this bad request exception? I think it is more like forbidden

Hm, actually they don't throw exception here if token is incorrect, so it's not ForbiddenException here

nmamro: > > Is this bad request exception? I think it is more like forbidden
> 
> Hm, actually they don't throw exception here if token is incorrect, so it's not ForbiddenException here

It's more like when you simply can't get connection from them for some reason. But I'm not sure that NotFound is good here

Clients/GUI/Scripts/BrokerManager.cs
nmamro: > Difficult to read. Split into 2 lines

Done

Clients/GUI/Scripts/BrokerManager.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public static class BrokerManager
+    {
+        public static async Task<IEnumerable<Instrument>> GetInstruments(
+                BankType bank,
+                string token,
+                string instrument,
+                int depth = 10)
+        {
+            const string url = "https://localhost:5003/brokers/instruments/get";
```
ifelseelif: Why it goes to broker service?

ifelseelif (changes requested):

Services/BrokerService/Controllers/BrokersController.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    [ApiController]
+    [Route("[controller]")]
+    public class BrokersController : ControllerBase
+    {
+        [Route("instruments/get")]
+        [HttpGet]
+        public IEnumerable<Instrument> GetInstruments(
+                [FromServices] IGetInstrumentsCommand command,
+                [FromQuery] BankType bank,
```
ifelseelif: Maybe replace those FromQuery on one object from body?

nmamro: > Maybe replace those FromQuery on one object from body?

That's a get request, it's said that sending GET requests with body is not a good practice

Clients/GUI/Scripts/BrokerManager.cs
nmamro: > Why it goes to broker service?

Fixing it next commits =)

Clients/GUI/Pages/Instruments.razor
dTarasova: Yeah but we don't use it anywhere else? And what is the difference: which properties we have to use as public and which as private

Common/DTO/BankBrokerObjects/Instrument.cs
dTarasova: Can you share link please?

Services/BrokerService/Commands/GetInstrumentsCommand.cs
dTarasova: Yes that was okay. But I think that we also decided to improve our code when we will have an opportunity. Also, In the unsorted I saw task create common interface for user command. We can use IBrokerCommand interface 
Or instrument. Don't know. But I think that a lot of different commands, consumers and etc. not very good practice

Services/BrokerService/Commands/GetInstrumentsCommand.cs
dTarasova: Because you will not log that you received request. You will catch only exception but the last positive log message would be from previous step

Clients/GUI/Pages/Instruments.razor
dTarasova: ok

Services/BrokerService/Commands/GetInstrumentsCommand.cs
nmamro: > Because you will not log that you received request. You will catch only exception but the last positive log message would be from previous step

It will enter try block anyway, why would it miss log?

Common/DTO/BankBrokerObjects/Instrument.cs
nmamro: > Can you share link please?

Just google "what is ticker"/"what is lot number" etc., these're common economic terms

nmamro: > Can you share link please?

May be I was wrong about explicit tinkoff docs about it, but the idea was that it's described in first google response

Services/BrokerService/Commands/GetInstrumentsCommand.cs
nmamro: > Yes that was okay. But I think that we also decided to improve our code when we will have an opportunity. Also, In the unsorted I saw task create common interface for user command. We can use IBrokerCommand interface
> Or instrument. Don't know. But I think that a lot of different commands, consumers and etc. not very good practice

Ok, agree, I will fix it :)

Clients/GUI/Pages/Instruments.razor
nmamro: > Yeah but we don't use it anywhere else? And what is the difference: which properties we have to use as public and which as private

Actually when it comes to razor page code I can't see any difference, but I made it private, no problem

Services/BrokerService/Commands/GetInstrumentsCommand.cs
nmamro: > Can we use just ICommand interface?

Added ICommand

Services/OperationService/Controllers/OperationsController.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.AspNetCore.Mvc;
+using DTO;
+using DTO.MarketBrokerObjects;
+using System.Threading.Tasks;
+using Interfaces;
+using DTO.BrokerRequests;
+
+namespace OperationService.Controllers
+{
+    [ApiController]
```
denis-kriv, line 11: Why do we use ApiControlller? Maybe we can use Controller?

Clients/GUI/Scripts/BrokerManager.cs
```diff
@@ -0,0 +1,33 @@
...
+                { "depth", depth.ToString() },
+                { "instrument", instrument.ToString() }
+            };
+
+            var client = new RestClient<object, IEnumerable<Instrument>>(url, RestRequestType.GET, queryParams: queryParams);
+
+            return await client.Execute();
+        }
+    }
+}
```
Maksim-Trolina, line 33: Perhaps we should leave an empty line at the end of the file (this problem is in almost every file of the extraction request).

dTarasova (commented):
Why do I can check instruments before I will sign in?

How about adding this page in the header to not changing URL each time. 

Also, what do you think about changing the count of elements that we can see on the page? I often choose big numbers in this situation. 

I think that GUI now seems a little dull. All is grey and white. I like grey colour and I like minimalism... But it is too minimal, I think.

Services/OperationService/Controllers/OperationsController.cs
Cerberous007: We use Api controllers in all services

Clients/GUI/Scripts/BrokerManager.cs
nmamro: Convince me why I have to spend time doing it for each existing file or suggest some extension to do it in one click

Integrations/CentralBankIntegrationLib/RussianCBInfo.cs
```diff
@@ -33,7 +34,7 @@ public List<ExchangeRate> GetCurrencies()
             var rates = new List<ExchangeRate>();
             for (int i = 0; i < data.Currencies.Count; i++)
             {
-                rates.Add(new ExchangeRate { Code = data[i].CharCode, Value = data[i].ValueInDigits });
+                rates.Add(new ExchangeRate { Code = data[i].CharCode, Value = decimal.Parse(data[i].ValueInString.Replace(',', '.')) });
```
dTarasova: I have an exception here. Why do you need to change this line
System.FormatException: 'Input string was not in a correct format.'

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
dTarasova: So?)

Services/OperationService/Controllers/OperationsController.cs
```diff
@@ -0,0 +1,34 @@
...
+    [Route("[controller]")]
+    public class OperationsController : ControllerBase
+    {
+        [Route("instruments/get")]
+        [HttpGet]
+        public async Task<IEnumerable<Instrument>> GetInstruments(
+                [FromServices] ICommand<GetInstrumentsRequest, IEnumerable<Instrument>> command,
+                [FromQuery] BrokerType broker,
```
dTarasova, line 19: Why do we use from query here? Isn't it would be easier to achieve some model from body here?

dTarasova (changes requested):
Have an exception in CBLib but all rest seems good for me

Services/OperationService/Controllers/OperationsController.cs
nmamro: > Why do we use from query here? Isn't it would be easier to achieve some model from body here?

Because it's GET

Integrations/TinkoffIntegrationLib/TinkoffBankBroker.cs
nmamro: > So?)

How should we log anything in library project?

Integrations/CentralBankIntegrationLib/RussianCBInfo.cs
nmamro: > I have an exception here. Why do you need to change this line
> System.FormatException: 'Input string was not in a correct format.'

There were some culture conflicts because my OS is in English I guess, I'll fix it so it works everywhere then

nmamro: > I have an exception here. Why do you need to change this line
> System.FormatException: 'Input string was not in a correct format.'

Fixed (i guess)

dTarasova (approved):

Cerberous007 (approved):

ifelseelif (approved):

Maksim-Trolina (approved):

PR #108, opened 2020-04-25
Added tests engine

nmamro:


GUITests/Tests/SignInTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using Kernel;
+using Kernel.Enums;
+using OpenQA.Selenium.Chrome;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace GUITestsEngine.Tests
+{
+    public class SignInTests
```
dTarasova, line 13: Why do sign in tests use create user request?

GUITests/Core/Engine.cs
```diff
@@ -0,0 +1,132 @@
...
+            });
+        }
+
+        private static void RunTestMethod(Type type, MethodInfo method)
+        {
+            var ctor = type.GetConstructor(Type.EmptyTypes);
+
+            if (ctor == null)
+            {
+                throw new FormatException($"{type.Name} must have parameterless constructor");
```
dTarasova, line 51: It is new custom exception?

nmamro: > It is new custom exception?

No

GUITests/Tests/SignInTests.cs
nmamro: > Why do sign in tests use create user request?

To be able to login user with correct data

GUITests/Core/Engine.cs
```diff
@@ -0,0 +1,132 @@
...
+            var instance = ctor.Invoke(null);
+
+            var isPassed = false;
+
+            try
+            {
+                method.Invoke(instance, null);
+                isPassed = true;
+            }
+            catch { }
```
denis-kriv, line 63: Why do we need this empty "catch block"?

GUITests/Core/Engine.cs
```diff
@@ -0,0 +1,132 @@
...
+            finally
+            {
+                testResults[type].Add(new TestInfo(method.Name, isPassed));
+            }
+        }
+
+        private static void PrintReport()
+        {
+            Console.WriteLine("Testing report:");
+            Console.WriteLine("-----------------------------");
```
denis-kriv, line 73: Why not to create a variable for this string?

denis-kriv (dismissed):

GUITests/Core/Engine.cs
denis-kriv: I meant this "-----------------------------"

GUITests/Tests/SignUpTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            Thread.Sleep(5000);
+
+            WebDriverWrapper.FindElementByClass(driver, "a", "Sign up").Click();
+
+            WebDriverWrapper.FillForm(driver, input);
+
+            WebDriverWrapper.FindElementByClass(driver, "button", "Next").Click();
+
+            Thread.Sleep(5000);
+        }
```
Maksim-Trolina, line 37: Bad practice. Better to use explicit expectations.

GUITests/Core/Engine.cs
nmamro: > Why not to create a variable for this string?

Why don't you like this string?

GUITests/Core/Engine.cs
nmamro: > Why do we need this empty "catch block"?

Not to fall after exception is thrown, but IsPassed was created ass false already, so we don't need to reassign its value here

GUITests/Core/Engine.cs
denis-kriv: You use this string too often and you can create a variable, but it is my opinion

dTarasova (approved):

dTarasova (approved):

Cerberous007 (approved):

Maksim-Trolina (approved):

ifelseelif (approved):

PR #115, opened 2020-05-06
Log Reader application

nmamro:


LogReader/LogReader/App.xaml.cs
```diff
@@ -0,0 +1,41 @@
...
+    public partial class App : Application
+    {
+        private readonly IHost host;
+
+        public App()
+        {
+            host = new HostBuilder()
+            .ConfigureServices((context, services) =>
+            {
+                services.AddScoped<LogViewModel>();
```
Cerberous007, line 20: Does AddScoped work correctly?

LogReader/LogReader/Models/Log.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+
+namespace LogReader.Models
```
Cerberous007: Path is LogReader/LogReader?

LogReader/LogReader/Models/Log.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class Log
```
Cerberous007: Why do you have 2 identical models with different names (LogMessage and Log)?

Services/AuthenticationService/Commands/LogoutCommand.cs
```diff
@@ -24,7 +24,7 @@ public bool Execute(Guid userId)
             if (userId == Guid.Empty)
             {
                 var e = new BadRequestException("Empty id in log out request.");
-                logger.LogWarning(e, "BadRequest thrown while trying to log out.");
+                logger.LogWarning(e, $"Attempt to logout user with empty id.");
```
Cerberous007: Why do you need $ here?

nmamro: > Why do you need $ here?

Don't need it here, thanks, will change in next commit

LogReader/LogReader/Models/Log.cs
nmamro: > Path is LogReader/LogReader?

I added a folder where I created the project, that's why there's such path

LogReader/LogReader/App.xaml.cs
nmamro: > Does AddScoped work correctly?

Why shouldn't it work?

LogReader/LogReader/Models/Log.cs
nmamro: > Why do you have 2 identical models with different names (LogMessage and Log)?

Log was automatically generated after database scaffolding, then didn't realize that it's better to use old LogMessage, changed it

.github/workflows/develop_build_actions.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:    
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
```
ifelseelif, line 12: What it is?

nmamro: > What it is?

Workflow setting which enables wpf apps build, Ubuntu does not support building wpf so it was crashing

Cerberous007 (approved):

strannick-rodion (approved):

VLiamin (approved):

ifelseelif (approved):

PR #116, opened 2020-05-08
Gui_FixedHeader

ifelseelif:


Clients/GUI/Components/Wizard.razor.cs
```diff
@@ -20,12 +20,6 @@ public partial class Wizard
         [Parameter]
         public string Id { get; set; }
 
-        /// <summary>
-        /// Function to check that there are no errors
-        /// </summary>
-        [Parameter]
-        public Func<bool> HasError { get; set; }
```
dTarasova, line 27: Why do we delete it?

Clients/GUI/Pages/SignUpWizzard.razor
```diff
@@ -132,37 +138,64 @@
...
         showLoading = true;
-        errorMessage = "";
+        
         var file = files[^1];
+
         if (file != null)
         {
-            fileName = file.Name;
             // 2MB =  2097152 byte
             if (file.Size > 2097152)
```
dTarasova: Magic constants :)

Clients/GUI/Shared/Header.razor
```diff
@@ -1,38 +1,90 @@
...
+            return;
+        }
+
+        var requestResult = await UserGetter.GetUserById(userToken);
+
+        if (requestResult.Avatar != null)
+            imgSrc = GetUserImage(requestResult.Avatar, requestResult.AvatarExtension);
+    }
+
+    private string GetUserImage(byte[] avatar, string extension)
```
dTarasova: Why don't we achieve just userAvatar model?

dTarasova (changes requested):
I think that this window is too small, some of emails wouldn't be fit there and it just seems like someone want to seems to thin. 

And I am not sure about window in window, but not sure. Probably, we can close first one?

I can click on my photo in the header , but nothing happens. How about also add a link to userInfo 
image

Clients/GUI/Pages/SignUpWizzard.razor
ifelseelif: the comment above explains everything

Clients/GUI/Components/Wizard.razor.cs
ifelseelif: I found a way easier

Clients/GUI/Shared/Header.razor
ifelseelif: added

Clients/GUI/Pages/SignUpWizzard.razor
ifelseelif: moved into constant

Clients/GUI/Shared/Header.razor
```diff
@@ -1,38 +1,90 @@
...
+
     private void SignOut()
     {
         Storage.ClearAsync();
+
         AuthenticationProvider.MarkSignedOut();
     }
-}
 
+    private async void LoadImageSrc()
```
nmamro, line 70: I suggest moving this logic to OnAfterRender, because now we can't get image on first page load because you get items from storage only after rendering the page. So you may create a variable for image and change it in onafterrender

nmamro (approved):
I like it a lot

Clients/GUI/Shared/Header.razor
nmamro: I might be wrong a bit in comment above
Another way is to notify the component that it has changed after loading img

nmamro: > I might be wrong a bit in comment above
> Another way is to notify the component that it has changed after loading img

Just add `StateHasChanged();` at the end of image loading method in component

dTarasova (approved):

Clients/GUI/Shared/Header.razor
ifelseelif: added

PR #117, opened 2020-05-10
SeverServices_UserOperationsWithMoney

ifelseelif:


Clients/GUI/Pages/UserTransactions.razor
```diff
@@ -0,0 +1,81 @@
+﻿@page "/transactions"
```
nmamro: Change the design so it's under menu **UserInfo  Transactions**, so the choice does not disappear after getting to transactions

nmamro (commented):
Ok

strannick-rodion (approved):

Cerberous007 (commented):
Do we need all this js files??

Services/DataBaseService/Repositories/TradeRepository.cs
```diff
@@ -166,5 +167,36 @@ public void UpdateUserBalance(UserBalance userBalance)
...
+                foreach (DbTransaction transaction in dbUserTransactions)
+                {
+                    userTransactions.Add(mapper.MapToTransaction(transaction));
+                }
+
+                return userTransactions;
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(e);
```
Cerberous007: Why not logging this exception?

Services/OperationService/Commands/GetUserTransactionsCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+        public async Task<IEnumerable<Transaction>> Execute(GetUserTransactionsRequest request)
+        {
+            var getUserTransactions = new GetUserTransactions
+            {
+                UserId = request.UserId
+            };
+
+            var result = await GetUserTransactions(getUserTransactions);
+
+            logger.LogInformation($"Transactions of user{request.UserId} taken successfully");
```
Cerberous007, line 47: If we throw exception in DataBase, this line also will be execute? correct me if I`m wrong

dTarasova: No, I think it wouldn't, it doesn't make full path of request

Cerberous007: Yes, you`re right

dTarasova (commented):
Seems good. Suggestions to add pagination and search by fields, but not necessary

Services/DataBaseService/Repositories/TradeRepository.cs
ifelseelif: it is my mistake, I forgot to delete

Clients/GUI/Pages/UserTransactions.razor
ifelseelif: Transactions moved in UserInfo

nmamro (approved):

denis-kriv (approved):

Cerberous007 (approved):

PR #126, opened 2020-05-20
GUI_UpdateBotsPage

dTarasova:


Clients/GUI/Pages/BotEdition.razor
```diff
@@ -0,0 +1,201 @@
...
+    private BotModel userInput;
+    private EditContext editContext;
+    private bool showLoading = false;
+    private string errorServerMessage = "";
+    private List<BotRuleModel> initialRules;
+    private UserToken userToken;
+
+    private void InitializeBotRules()
+    {
+        BotRuleModel ruleBuy = null;
```
nmamro, line 88: Why not to unite buy and sell options in one rule?

nmamro (approved):

Clients/GUI/Pages/BotEdition.razor
dTarasova: They have little different descriptions. And I think that it could be important to have the opportunity for person to select two different parameters for sell and buy

PR #127, opened 2020-05-23
UserService_EmailSender_AddLogsAndException

strannick-rodion:


Services/UserService/Utils/EmailSender.cs
```diff
@@ -33,19 +34,25 @@ public void SendEmail(string email, ISecretTokenEngine secretTokenEngine)
...
                 }
-                catch (SmtpException)
+                catch (SmtpException e)
                 {
-                    Thread.Sleep(10000 * k);
+                    Thread.Sleep(10000 * (++i));
                     flag = false;
-                    if (k < 5)
-                        k++;
+                    logger.LogWarning(e, "SmtpException thrown while trying to Send Eamil to confirm");
```
nmamro: It's not formatted for correlation

Services/UserService/Utils/EmailSender.cs
```diff
@@ -21,7 +22,7 @@ public void SendEmail(string email, ISecretTokenEngine secretTokenEngine)
             var from = new MailAddress("t.platform@mail.ru", "Trading Station");
             var to = new MailAddress(email);
             string secretToken = secretTokenEngine.GetToken(email).ToString();
-            string link = $"https://localhost:44335//confirm/{secretToken}";
+            string link = $"https://localhost:44335/confirm/{secretToken}";
```
denis-kriv, line 25: We will use http

Services/UserService/Utils/EmailSender.cs
```diff
@@ -33,19 +34,25 @@ public void SendEmail(string email, ISecretTokenEngine secretTokenEngine)
...
                 {
                     flag = true;
                     smtp.Send(m);
+                    break;
                 }
-                catch (SmtpException)
+                catch (SmtpException e)
                 {
-                    Thread.Sleep(10000 * k);
+                    Thread.Sleep(10000 * (++i));
```
denis-kriv, line 47: why not to use `Thread.Sleep(10000)`

Services/UserService/Utils/EmailSender.cs
```diff
@@ -33,19 +34,25 @@ public void SendEmail(string email, ISecretTokenEngine secretTokenEngine)
 
             var flag=true;
```
denis-kriv, line 35: spaces

denis-kriv: And do we need flag?

Services/UserService/Utils/EmailSender.cs
strannick-rodion: Because we wont to have more time after every try

Cerberous007 (approved):

PR #128, opened 2020-05-25
Bots logic

nmamro:


Services/OperationService/Commands/RunBotCommand.cs
```diff
@@ -1,46 +1,79 @@
...
             {
                 var e = new NotFoundException("Not found bot to run");
-                logger.LogWarning(e, $"{e.Message}, botId: {request.ID}");
+                logger.LogWarning(e, $"{e.Message}, botId: {runRequest.BotId}");
+                throw e;
+            }
+            catch (BadRequestException e)
+            {
+                logger.LogWarning(e, $"{e.Message}, botId: {runRequest.BotId}");
                 throw e;
```
denis-kriv, line 77: maybe we should catch exceptions of all types

Services/OperationService/Bots/BotRules/BotRule.cs
```diff
@@ -0,0 +1,132 @@
...
+                    );
+
+                    trigger.Triggered += async (s, e) =>
+                    {
+                        await Execute(s, e);
+                    };
+
+                    triggers.Add(trigger);
+                }
+                catch
```
denis-kriv, line 122: May be `catch(Exception)`

nmamro: > May be `catch(Exception)`

The difference between catch and catch(Exception) is that when we write just catch, it catches all types of possible excpetions even if they are not present in .net, and actually in our case it does not make any sense, but agree, mb it's better because of common styleguide :)

Services/OperationService/Commands/RunBotCommand.cs
nmamro: > maybe we should catch exceptions of all types

Yeah, there's some reason in adding default handler mb

strannick-rodion (approved):

